PR #24, opened 2026-02-16
fix(cli): add missing shutdown_plugins() in astridd shutdown path

joshuajbouw:
## Summary

- The `astridd` binary (`daemon_main.rs`) called `shutdown_servers()` but not `shutdown_plugins()` before shutting down
- This could leave MCP plugin child processes running as orphans after daemon shutdown
- Adds the missing `shutdown_plugins().await` call, matching the shutdown sequence already used in `commands/daemon.rs`

## Test plan

- [x] `cargo check -p astrid-cli` — clean
- [x] Manual: start daemon with MCP plugins loaded, stop daemon, verify child processes are terminated

### DIFF
--- a/crates/astrid-cli/src/daemon_main.rs
+++ b/crates/astrid-cli/src/daemon_main.rs
@@ -97,6 +97,10 @@ async fn main() -> Result<()> {
         h.abort();
     }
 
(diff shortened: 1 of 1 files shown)


joshuajbouw:
Can confirm that child processes are shut down.

PR #25, opened 2026-02-16
fix(gateway): reduce plugin registry lock duration during load/unload

joshuajbouw:
## Summary

- The write lock on the plugin registry was held across expensive async operations: `plugin.load()` (MCP handshake + subprocess spawn) and `plugin.unload()` (child process teardown)
- This blocked all other registry readers/writers for the entire duration — potentially seconds for MCP plugins
- Restructured to a take-out/put-back pattern: briefly lock to `unregister` the plugin, perform the async operation without the lock, then briefly lock again to `register` it back
- Applied to: RPC `load_plugin`, RPC `unload_plugin`, and the startup auto-load background task

## Test plan

- [x] `cargo check -p astrid-gateway -p astrid-cli` — clean
- [x] `cargo clippy` — no errors
- [x] `cargo fmt --check` — no issues
- [x] Manual: load an MCP plugin while simultaneously listing plugins — verify no blocking

### DIFF
--- a/crates/astrid-gateway/src/server.rs
+++ b/crates/astrid-gateway/src/server.rs
@@ -397,33 +397,55 @@ impl DaemonServer {
 
         // Background task: auto-load discovered plugins after the server is
         // accepting connections (avoids blocking the 5s CLI connect timeout).
+        // Each plugin is taken out of the registry, loaded without the lock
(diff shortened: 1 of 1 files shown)


joshuajbouw:
Confirmed not blocking when running a plugin install and a plugin list at the same time with two different terminals.

PR #440, opened 2026-03-14
feat(capsule-fs): add create_directory, delete_file, and move_file tools

### LINKED ISSUE #416: Native Filesystem Lifecycle Tools
`astrid-capsule-fs` is missing basic lifecycle tools. Because these are absent, the agent must fall back to using `run_shell_command("mkdir")`, which repeatedly triggers human-in-the-loop security approvals for mundane operations.

### Tasks
- Natively implement `mkdir`, `delete_file`, and `move_file` tools in `astrid-capsule-fs`.

joshuajbouw:
## Summary
- Adds `create_directory`, `delete_file`, and `move_file` tools to `astrid-capsule-fs` so agents stop falling back to `run_shell_command` for mundane filesystem lifecycle operations
- `move_file` uses read-write-delete pattern (no VFS rename primitive) with 10MB size guard and destination rollback on source removal failure
- `delete_file` validates target is a file (not directory) and exists before attempting removal

## Known Limitations
- `delete_file` and `move_file` only work on files created during the current session (upper layer). Workspace files (lower layer) return `PermissionDenied` because `OverlayVfs` lacks whiteout support. Clear error messages inform the agent of this constraint.
- `move_file` is non-atomic (copy+unlink). A native `fs::rename` syscall would be the proper long-term solution.

## Test Plan
- `cargo test --workspace -- --quiet` (all 1285 tests pass)
- `cargo clippy -- -D warnings` (clean)
- `cargo fmt --check` (clean)
- Internal review: 2 rounds, all findings addressed
- Gemini review: 5 findings posted, all resolved after discussion

## Related Issues
Closes #416

### DIFF
--- a/crates/astrid-capsule-fs/src/lib.rs
+++ b/crates/astrid-capsule-fs/src/lib.rs
@@ -159,13 +159,16 @@ impl FsTools {
 
     #[astrid::tool("delete_file")]
     pub fn delete_file(&self, args: DeleteFileArgs) -> Result<String, SysError> {
-        if !fs::exists(&args.file_path)? {
-            return Err(SysError::ApiError(format!(
-                "file does not exist: {}",
-                args.file_path
-            )));
-        }
-        if file_stat(&args.file_path)?.is_dir {
+        let stat = match file_stat(&args.file_path) {
+            Ok(s) => s,
+            Err(_) => {
+                return Err(SysError::ApiError(format!(
+                    "file does not exist: {}",
+                    args.file_path
+                )));
+            }
+        };
+        if stat.is_dir {
             return Err(SysError::ApiError(format!(
                 "{} is a directory, not a file; delete_file only supports files",
                 args.file_path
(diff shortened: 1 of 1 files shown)


joshuajbouw:
Addressed both findings from @gemini-code-assist in d9826de8:

**Re: `file_stat` `unwrap_or` defaults ([comment](https://github.com/unicity-astrid/astrid/pull/440#discussion_r2935584691))** - Agreed. `file_stat` now returns `SysError::ApiError` when `isDir` or `size` fields are missing or invalid, instead of silently defaulting to `false`/`0`. Missing required data from an FFI boundary should be an error.

**Re: redundant `fs::exists` in `delete_file` ([comment](https://github.com/unicity-astrid/astrid/pull/440#discussion_r2935584694))** - Agreed. Removed the extra `fs::exists` check and now use `file_stat` error to detect missing files, matching the pattern already used in `move_file`. One fewer host round-trip.

PR #711, opened 2026-05-13
fix(capsule): treat wasip2 stub run() exports as no-run-loop

### LINKED ISSUE #710: bug(capsule): wasip2 stub run() exports misclassified as live run-loops
## Summary

After the wasmtime Component Model migration, the kernel's run-loop prescan misclassifies every Component Model capsule as a live run-loop daemon. The Component Model emits a `run()` export for the WASI preview-2 ABI even when the capsule has no actual run loop — the export is a stub that calls `wasi:cli/exit` and never blocks. The kernel sees the export name and treats every capsule as a daemon that needs the 100ms post-load detection grace period, then keeps a background task alive waiting for it to exit. Stateless interceptor-only capsules pay the cost of a never-terminating monitor task each.

## Motivation

Every Component Model capsule we ship hits this — the kernel boots with 15 phantom run-loop monitors that never fire. It's not a crash, but it's a measurable resource overhead (one tokio task + one monitor channel per capsule for the lifetime of the daemon) and the prescan log spam ("Capsule X has a run loop, monitoring") is misleading during incident triage.

## Proposed implementation

Detect the wasip2 stub shape by looking at the import dependency set of the `run()` export: a real run loop imports `astrid_ipc::recv` / `astrid_ipc::poll` and the bus subscribe handle table; a stub only imports `wasi:cli/exit`. The prescan already walks the component's import graph for the IPC-host-fn check; extend that walk to also classify `run()` as stub-or-real based on which hosts it transitively depends on.

For ambiguity cases (capsule has both `run()` AND `install()` / `upgrade()` exports — the Component Model trio), apply the same imports check on each in sequence. A capsule with three stub exports gets all three classified as stubs and no monitor task starts.

## Verification

- 6 new unit tests in `crates/astrid-capsule/src/engine/wasm/mod.rs`:
  - stub aliasing (all three exports same instance)
  - distinct-index real runs
  - all-three-real
  - real-install-with-stub-run
  - lifecycle stubs
  - non-trio names
- Pre-existing prescan tests unchanged and still pass.
- Manual against a daemon load — the 15 Component Model capsules in the standard distro stop logging "monitoring run loop"; real run-loop daemons (cli, context-engine, prompt-builder, registry) keep using the auto-subscribe path from #343.

joshuajbouw:
## Linked Issue

Closes #710.

## Summary

After the wasmtime Component Model migration, the kernel's run-loop prescan misclassified every Component Model capsule as a live run-loop daemon. The Component Model emits a `run()` export for the WASI preview-2 ABI even when the capsule has no actual run loop — the export is a stub that calls `wasi:cli/exit` and never blocks. The kernel was treating every capsule as a daemon, paying the cost of a never-terminating monitor task per capsule, and emitting misleading "monitoring run loop" log spam during incident triage.

## Changes

* `crates/astrid-capsule/src/engine/wasm/mod.rs`:
  - Prescan now walks the `run()` export's transitive import set to classify it as a stub vs a real run loop. A real run loop imports `astrid_ipc::recv` / `astrid_ipc::poll` and the bus subscribe handle table; a stub only imports `wasi:cli/exit`.
  - For the Component Model trio (`run()` + `install()` + `upgrade()`), the same check applies per export — three stub exports gets all three classified as stubs and no monitor task starts.
  - Six new tests cover stub aliasing, distinct-index real runs, all-three-real, real-install-with-stub-run, lifecycle stubs, and non-trio names.

## Test Plan

- [x] `cargo clippy -p astrid-capsule --no-deps -- -D warnings`
- [x] `cargo test -p astrid-capsule --lib` (278 passed, includes the 6 new prescan tests)
- [x] Manual: standard distro boot — 15 Component Model capsules stop logging "monitoring run loop"; real run-loop daemons (cli, context-engine, prompt-builder, registry) keep using the auto-subscribe path from #343.

## Checklist

- [x] Linked to an issue (#710)


joshuajbouw:
Addressed in 6aa1c0f: added `ComponentExternalKind::Func` filter to the component-export pre-scan so we only compare indices within the function index space (matches the existing module-export scan above). Did not switch parse-error handling to fail-fast — that contradicts the documented fail-open contract ("on any parse error, return true"), which is intentional to avoid wedging boot on unknown wasm shapes. Also rebased on origin/main to resolve the CHANGELOG conflict.

PR #713, opened 2026-05-13
feat(manifest): Cargo-shaped Capsule.toml schema with [publish] / [subscribe] tables

### LINKED ISSUE #712: feat(manifest): Cargo-shaped Capsule.toml schema with [publish] / [subscribe] tables
## Summary

Add a Cargo-like alternative for declaring topic ACLs and capabilities in `Capsule.toml`. The legacy form (`[[topic]]` blocks + `[capabilities].ipc_publish`/`ipc_subscribe` lists) keeps working unchanged; the new form (`[publish]` / `[subscribe]` tables, one row per topic, with the WIT envelope reference and ACL declaration colocated) is what RFCs are converging on for the SDK / contracts work.

The RFC text lives in `astrid-rfcs/cargo-like-manifest`. This PR is the parser-side enablement: both surfaces parse into the same `CapsuleManifest`; capsule authors pick the form that reads cleanest for their case during the migration window. The legacy form will eventually be deprecated once every shipping capsule has been ported.

## Motivation

The current schema spreads a single topic's information across three places (`[[topic]]` for the WIT envelope, `[capabilities].ipc_publish` for the publish allowlist, `[capabilities].ipc_subscribe` for the subscribe allowlist). For capsules with 10+ topics that gets noisy fast and makes it easy to forget the ACL entry when adding a topic. The Cargo-shaped form colocates the envelope and ACL in a single row per topic, matching how Cargo dependencies, features, and `[dependencies]` already shape Rust manifests — operators already know the pattern, no new shape to learn.

\`\`\`toml
[publish]
"agent.v1.response" = { wit = "@unicity-astrid/wit/agent/response" }
"client.v1.disconnect" = { wit = "TODO" }

[subscribe]
"astrid.v1.capsules_loaded" = { wit = "TODO" }
"astrid.v1.elicit.*" = { wit = "TODO" }
\`\`\`

Equivalent legacy form:

\`\`\`toml
[[topic]]
name = "agent.v1.response"
wit = "@unicity-astrid/wit/agent/response"

[[topic]]
name = "client.v1.disconnect"
wit = "TODO"

[capabilities]
ipc_publish = ["agent.v1.response", "client.v1.disconnect"]
ipc_subscribe = ["astrid.v1.capsules_loaded", "astrid.v1.elicit.*"]
\`\`\`

## Proposed implementation

`crates/astrid-capsule/src/manifest.rs`:

* New `PublishEntry` / `SubscribeEntry` structs, deserialized from the table rows. Each carries the `wit` string and any future ACL knobs (priority, audit hints).
* `CapsuleManifest` gains `publish: HashMap<String, PublishEntry>` and `subscribe: HashMap<String, SubscribeEntry>` fields. Both are `#[serde(default)]` so legacy manifests parse unchanged.
* A post-deserialize normalize pass merges the new tables into the existing `topics` / `capabilities.ipc_publish` / `capabilities.ipc_subscribe` data so downstream consumers (kernel ACL check, prescan, install validator) need zero changes.
* The two forms cannot describe the same topic — declaring `agent.v1.response` in both `[publish]` and `[[topic]]` is a parse error, so manifests don't accidentally drift across forms.

Validation surface and error messages match the legacy form's: bad WIT ref, unknown segment, empty topic name, leading/trailing-dot, mid-segment wildcard.

## Verification

- Unit tests in `manifest.rs` for: legacy-only parses unchanged; new-only parses; mixed disjoint topics merge cleanly; same topic in both forms rejected; per-form ACL-derivation correctness.
- All 16 shipping capsules' manifests still parse after the change (the parser test suite reads them via `include_str!`).
- `cargo clippy --workspace --all-features -- -D warnings` clean.

## Alternatives considered

- **Convert all capsules at once to the new form** — too disruptive; we ship 16 capsules in three different repos and want gradual migration. Keep both forms.
- **Auto-derive `topics` + `capabilities.ipc_*` purely from `[publish]` / `[subscribe]`** — works for new capsules but breaks the smooth migration story for legacy manifests that still want to add a single new topic.

## Dependencies

- **RFC text:** `astrid-rfcs/cargo-like-manifest` (already merged on the RFC repo).
- **Migration follow-up:** porting the 16 shipping capsules to the new form is per-capsule churn, not blocking on this PR.

joshuajbouw:
## Linked Issue

Closes #712.

## Summary

Adds a Cargo-shaped alternative for declaring topic ACLs and capabilities in `Capsule.toml`. The legacy form (`[[topic]]` blocks + `[capabilities].ipc_publish`/`ipc_subscribe` lists) keeps working unchanged; the new form (`[publish]` / `[subscribe]` tables, one row per topic with the WIT envelope reference and ACL declaration colocated) is what the RFC work is converging on. RFC text lives in `astrid-rfcs/cargo-like-manifest`.

Both surfaces parse into the same `CapsuleManifest` — downstream consumers (kernel ACL check, prescan, install validator) need zero changes. Capsule authors pick the form that reads cleanest during the migration window; legacy will be deprecated once every shipping capsule has been ported.

```toml
[publish]
"agent.v1.response" = { wit = "@unicity-astrid/wit/agent/response" }
"client.v1.disconnect" = { wit = "TODO" }

[subscribe]
"astrid.v1.capsules_loaded" = { wit = "TODO" }
"astrid.v1.elicit.*" = { wit = "TODO" }
```

## Changes

* `crates/astrid-capsule/src/manifest.rs`:
  - New `PublishDef` / `SubscribeDef` structs with custom `Deserialize` accepting either bare-string short form or full-table long form.
  - `CapsuleManifest` gains `publishes: HashMap<String, PublishDef>` (TOML key `publish`) and `subscribes: HashMap<String, SubscribeDef>` (TOML key `subscribe`), both `#[serde(default)]`.
  - `effective_ipc_publish_patterns()` / `effective_ipc_subscribe_patterns()` / `effective_interceptors()` merge the new tables with legacy `[capabilities].ipc_publish` / `ipc_subscribe` / `[[interceptor]]` — single source of truth downstream.
  - **Deserialize-time validator**: a `[publish]` or `[subscribe]` entry that sets more than one of `version` / `tag` / `rev` / `branch` / `path` is rejected with a typed `serde::de::Error::custom` carrying the offending topic key. Catches ambiguous source pinning at install-time instead of surfacing as inconsistent resolver behaviour later.
* `CHANGELOG.md` — extended the `### Added` cargo-like-manifest bullet to mention the source-pin validator.
* Tests added (`crates/astrid-capsule/src/manifest.rs#tests`): publish/subscribe short form, long form with zero pins, long form with one pin, long form with two pins rejected (with the error message asserting the topic key appears in the error).

## Test Plan

- [x] `cargo clippy -p astrid-capsule --no-deps -- -D warnings` clean
- [x] `cargo test -p astrid-capsule --lib` (279 passed, including 7 new validator tests)
- [x] All 16 shipping capsule manifests in the parser fixture suite still parse cleanly after the change.

## Checklist

- [x] Linked to an issue (#712)


joshuajbouw:
Addressed in 5c4897b5ddf5c0056529ac239d983a7bccfc416b.

**Suggestion 1 (`SourcePin` enum):** valid on principle — making invalid states unrepresentable is the right Rust idiom when there's a consumer that benefits from matching on the enum. Today, zero code in the workspace reads these fields (grepped across `core/`), so the type-level benefit has no current beneficiary. The runtime invariant is enforced at the parse boundary with a deserialize-time validator instead, which closes the wire-input side. When the resolver / lockfile work lands and starts matching on the source pin, switching to an enum is a small, motivated refactor done by that author at that moment. Not paying for it now.

Note: the cargo-like-manifest RFC is still in draft. The RFC currently says "exactly one" but the existing short-form deserializer produces zero pins, so the implemented invariant is "at most one." Will reconcile the RFC text before it lands — either tighten the code to require one, or relax the RFC to match the code. Flagging here so it doesn't get lost.

**Suggestion 2 (DRY the Deserialize impls):** declined. The two impls share visual shape but the `LongForm` structs they wrap differ semantically — `PublishDef` carries `fanout: bool`, `SubscribeDef` carries `handler: Option<String>`. Factoring this out requires either a generic over the extra field or a macro, and the resulting indirection reads worse than two short, explicit impls. The duplication is local and won't grow.

PR #715, opened 2026-05-13
feat(manifest): EnvScope { Agent, Shared } operator-only sharing model for env_type=secret

### LINKED ISSUE #714: feat(manifest): EnvScope { Agent, Shared } operator-only sharing model for env_type=secret
## Summary

Groundwork for the secret-store routing follow-up. Adds an `EnvScope { Agent, Shared }` enum to `astrid-capsule::manifest::EnvDef` so the operator's `secret set --scope agent|shared` flag has a type to plumb through; the operator decides scope, the capsule does NOT — capsules come from external sources and can't be trusted to mark their own credentials as host-shared.

## Motivation

When the secret-store fix lands (separate PR), the kernel will route `env_type = "secret"` values to a sharing-aware backend. The sharing decision needs an operator-facing knob, and clap's `ValueEnum` derivation wants a real enum at the manifest layer to point at. This PR introduces the enum and its `is_default()` helper without yet changing any code path — additive only — so the actual routing change can land separately and keep its diff focused.

Critical security property documented on the enum: an external capsule must NOT be able to declare `EnvScope::Shared` on its own secrets. If it could, a malicious Telegram-bot capsule could mark its bot token "shared" and every agent on the host would inherit it. The doc-comment makes the operator-only contract explicit so future code changes don't accidentally introduce a capsule-side override path.

## Proposed implementation

\`\`\`rust
#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize)]
#[serde(rename_all = "lowercase")]
pub enum EnvScope {
    /// Per-invoking-principal storage. Stored under
    /// `astrid:secret:<principal>:<capsule>` and consulted first on
    /// every per-invocation read. Fail-closed default at the CLI.
    #[default]
    Agent,
    /// Host-wide storage under `astrid:secret:__host__:<capsule>`.
    /// The kernel-side read path falls through to this slot when a
    /// per-agent lookup misses.
    Shared,
}
\`\`\`

Capsule manifests do NOT declare scope. The `EnvDef` shape is unchanged; the enum is referenced from the CLI's `--scope` flag (separate PR for the routing change).

## Verification

- `cargo build -p astrid-capsule`
- `cargo clippy -p astrid-capsule --no-deps -- -D warnings`
- Additive change — no behaviour to verify yet; the contract documentation is the load-bearing piece.

joshuajbouw:
## Linked Issue

Closes #714.

## Summary

Additive enum on `astrid-capsule::manifest::EnvDef` so the operator's `secret set --scope agent|shared` flag (separate PR) has a type to plumb through. The doc-comment locks in the **operator-only contract**: capsule manifests cannot declare scope themselves because they come from external sources — a malicious capsule could otherwise mark its own bot token "shared" and inherit every agent's invocation budget.

The contract is enforced at the parser via `#[serde(skip_deserializing)]` on the field, so a manifest `scope = "shared"` declaration is silently ignored and falls back to the `Agent` default. The operator can still construct `EnvScope::Shared` in Rust at runtime (e.g. from `secret set --scope shared`), which serializes correctly — only the *manifest input path* is blocked.

## Changes

- `crates/astrid-capsule/src/manifest.rs`:
  - New `EnvScope { Agent, Shared }` enum (lowercase-serialized) with `Agent` as the `Default`. `is_default()` helper so serializers can omit the field by convention.
  - `EnvDef.scope: EnvScope` field with `#[serde(default, skip_serializing_if = "EnvScope::is_default", skip_deserializing)]` enforcing the operator-only contract.
  - Doc fixes: `env_type` example list now enumerates the real accepted types (`text`, `secret`, `select`, `array`) instead of the obsolete `string`/`choice`; `EnvScope::Shared` doc cites the correct flag (`--scope shared`, not `--scope host`).
- `crates/astrid-capsule/tests/env_scope_manifest_isolation.rs` (new):
  - Pinning test that a manifest with `scope = "shared"` deserializes back to `EnvScope::Agent`.
  - Round-trip test confirming operator-constructed `Shared` serializes to `scope = "shared"`.
- `crates/astrid-capsule/src/engine/wasm/mod.rs` and `crates/astrid-integration-tests/tests/wasm_env_e2e.rs`: existing EnvDef literals in unit and integration tests gain the new `scope` field with `EnvScope::default()`.
- `CHANGELOG.md`: new `### Added` bullet under `[Unreleased]`.

## Test Plan

- [x] `cargo clippy --workspace --all-features -- -D warnings` clean (matches CI)
- [x] `cargo test -p astrid-capsule` (2 new tests pass; pre-existing tests unchanged)
- [x] `manifest.rs` is 999 lines — under the 1000-line CI threshold

## Checklist

- [x] Linked to an issue (#714)


joshuajbouw:
Addressed in 9ac5d6209347d80897bcbc904f8dd5b7ad12a7b7 (rebased on origin/main; CHANGELOG conflict resolved).

**Suggestion 1 (security-high — `#[serde(skip_deserializing)]` on `scope`):** valid and important. The CHANGELOG entry already commits to "the capsule manifest does NOT declare scope" but the implementation didn't enforce it. Applied verbatim. Added a pinning test at `crates/astrid-capsule/tests/env_scope_manifest_isolation.rs` that asserts a manifest with `scope = "shared"` still deserializes with `scope = EnvScope::Agent`. The companion test confirms round-trip serialization works when the operator constructs the value in Rust (i.e. the kernel can still write Shared into the manifest representation programmatically; it just won't read one from disk).

**Suggestion 2 (env_type doc accurate types):** valid. The doc listed `"string"` and `"choice"`, but `crates/astrid-capsule/src/engine/wasm/host/elicit.rs:21-32` matches on `"text" | "secret" | "select" | "array"`. Doc rewritten to enumerate the actual accepted set with a note on what each does (`select` uses `enum_values`, `array` is comma-separated).

**Suggestion 3 (`--scope host` → `--scope shared`):** valid. Editorial slip from a draft where the variant was named `Host`; the enum renames to lowercase so the actual flag value is `shared`. Fixed.

Note on file size: `manifest.rs` is sitting at 999 lines after this change. The new tests went into a separate integration test file rather than expanding the inline module — both to land under the 1000-line CI threshold and because the security-isolation test is contract-pinning, not the kind of "is this parser branch correct" sanity check that belongs co-located.

PR #717, opened 2026-05-13
fix(multitenancy): provision agent home, drop --agent on capsule cmds

### LINKED ISSUE #716: bug(multitenancy): agent.create skips home tree provisioning + capsule cmds wrongly take --agent
## Summary

Two related findings from the multi-tenancy live-test (F-I and F-J in the local sysadmin report):

1. **`agent.create` does not provision the new principal's home tree.** A freshly created agent has a `profile.toml` and an identity-store entry, but `~/.astrid/home/{principal}/.local/{kv,log,audit,tokens,tmp,capsules}` are missing. The first per-invocation IPC for that principal hits the per-principal scopes (`HostState::install_*`) and silently falls back to the capsule owner's directories — which means session/turn data Alice writes lands in default's KV namespace, audit entries route to default's audit log, etc. Per-principal isolation collapses for any principal not provisioned through the legacy distro-apply path.

2. **Capsule lifecycle subcommands accept `--agent`** even though capsules are shared across principals (NEAR-style: one shared contract, per-invocation principal scoping). `astrid capsule install --agent alice X` looks like it installs `X` for alice only, but the kernel installs the capsule globally and ignores the flag. Operators who used the flag thinking it gave them per-agent capsule sets carry a wrong mental model into every later debugging step.

## Motivation

Both issues actively mislead operators during multi-tenant rollout: (1) silently routes Alice's chat history into default's namespace (a confidentiality boundary failure that won't trip any error path), and (2) makes `astrid capsule list --agent alice` return *every* capsule on the host, conflicting with the operator's per-agent mental model.

### F-J — `agent.create` provisions home tree

Have `agent_create` in `crates/astrid-kernel/src/kernel_router/admin/handlers.rs` call `kernel.astrid_home.principal_home(&principal).ensure()` immediately after the identity-store link succeeds and before returning success. `PrincipalHome::ensure()` already creates every per-principal subdirectory with the right `0o700` perms — same code path the bootstrap seeds default's home with.

Best-effort: a directory failure logs at `warn` with `security_event = true` and does NOT roll back the profile. The principal still exists; the next IPC will re-attempt creation lazily.

### F-I — drop `--agent` from capsule cmds

Remove the `--agent` flag from `astrid capsule install / list / remove / update / config / show` in `crates/astrid-cli/src/commands/capsule/`. Capsules are global. Operators who actually want per-principal capsule scoping use `secret set --agent <name>` to give the same shared capsule different per-principal env values; the capsule WASM stays one instance.

Help text update: capsule subcommand docs note that capsules are shared across principals and per-agent customization happens via `secret set` / future `quota set --agent`.

## Verification

- New `agent_create_provisions_home_tree` integration test asserts `~/.astrid/home/alice/.local/kv` exists after `agent.create alice`.
- Manual: `agent create alice && astrid -p "..."` (as alice) writes session data under `~/.astrid/home/alice/.local/kv/`, not under default's.
- `astrid capsule install` without `--agent` still works against the kernel; help text + completions reflect the removal.

joshuajbouw:
## Linked Issue

Closes #716.

## Summary

Two related multi-tenancy fixes from the local sysadmin live-test report (F-I + F-J):

* **F-J — `agent.create` provisions the per-principal home tree, fail-closed.** Calls `kernel.astrid_home.principal_home(&principal).ensure()` after the identity-store link succeeds. Per-principal KV, log, audit, tokens, tmp, capsules dirs all materialize with `0o700` perms. Without this, the first per-invocation IPC for the new principal silently falls back to the capsule owner's directories — Alice's chat history would land in default's KV namespace, breaking the confidentiality boundary with no error path. If provisioning fails, the handler rolls back the identity link and profile file rather than leaving a half-provisioned agent whose future invocations would leak into another principal's namespace.
* **F-I — `astrid capsule install / list / remove / update` no longer take `--agent`/`--group`.** Capsules are shared across principals (NEAR-style: one shared contract, per-invocation principal scoping). The flag was a misleading no-op — the kernel ignored it and installed/listed/removed globally. Help text now documents the shared-capsule model and points operators at `astrid capsule config -a <agent>` and `astrid secret set ... -a <agent>` for the per-tenant data surface that still exists.

## Changes

* `crates/astrid-kernel/src/kernel_router/admin/handlers.rs`: `agent_create` calls `kernel.astrid_home.principal_home(&principal).ensure()` after the identity-store link. **On failure, rolls back: unlinks the identity, deletes the user record, removes the profile file**, then returns an internal error. Previously the failure was logged as a warning and execution continued — that left the agent half-provisioned and downstream lookups would silently inherit default's namespace.
* `crates/astrid-cli/src/commands/capsule/`: drop `--agent` / `--group` from `install / update / list / remove`. Help text documents the shared-capsule model.
* `crates/astrid-kernel/src/kernel_router/admin/state_tests.rs`:
  - Existing `agent_create_writes_profile_and_links_identity` already asserts the home tree exists post-create.
  - New `agent_create_rolls_back_when_home_provisioning_fails` plants a regular file at the expected home directory path so `ensure()` fails, then asserts the identity link is gone, the profile file is gone, and the original blocker file is untouched.

## Test Plan

- [x] `cargo clippy --workspace --all-features -- -D warnings`
- [x] `cargo test -p astrid-kernel --lib agent_create` (5 passed, includes the new rollback test)
- [x] Manual: `agent create alice` produces `~/.astrid/home/alice/.local/{kv,log,audit,tokens,tmp,capsules}` and `~/.astrid/home/alice/.config/env`; `astrid capsule install` works without `--agent` and the flag is gone from `--help`.

## Checklist

- [x] Linked to an issue (#716)
- [x] CHANGELOG.md updated — one `### Changed` bullet, two `### Fixed` bullets under `[Unreleased]`


joshuajbouw:
Addressed in 50b4a0f34287a035ad8aa42f1852f3377fae734d (rebased on origin/main; CHANGELOG cleaned).

**Suggestion 1 (security-critical — home tree provisioning is non-fatal):** valid and applied. The PR description was right that downstream lookups silently fall back to the default principal's namespace on missing home tree — a real confidentiality break across tenants. The handler now treats home-tree-creation failure as a hard error and rolls back identity link + profile file before returning. The original "best-effort" rationale (lazy recovery) didn't hold up: the per-invocation host setup doesn't retry, it just resolves to whatever directory exists, which means another principal's. Better to leave nothing on disk than to leave a leaky agent.

Pinning test at `crates/astrid-kernel/src/kernel_router/admin/state_tests.rs::agent_create_rolls_back_when_home_provisioning_fails` forces `ensure()` failure by planting a regular file at the expected directory path, then asserts the identity link is gone, the profile file is gone, and the original blocker file is untouched.

**Suggestion 2 (CHANGELOG had unrelated entries):** valid — was caused by a stale rebase that pre-dated #709 / #711 / #713 / #715 landing on main. Fully rewritten by rebasing onto current main and adding only the entries this PR actually introduces: one `### Changed` (capsule cmds drop --agent), two `### Fixed` (agent.create fail-closed; read_message broadcast tolerance). Diff against main is now exactly the three new bullets.

Issue #658, opened 2026-04-21
feat: remote CLI authentication — ed25519 keypairs, device flow, and machine tokens


joshuajbouw:
## Summary

When the Astrid daemon serves a TCP/TLS endpoint for remote CLI connections, users need to authenticate to prove which agent they are. The local CLI uses the Unix socket + session token (file permissions = identity). Remote CLI needs an explicit auth mechanism.

Three methods, each for a different scenario. Ed25519 keypairs are the primary method — already native to Astrid's cryptographic infrastructure.

### 1. Ed25519 Keypairs (primary, Astrid-native)

Astrid already uses ed25519 for runtime keys, capability token signing, audit chain integrity, and Agent Card signatures. Extend this to user authentication via challenge-response.

**Provisioning flow:**

```bash
# Admin generates a keypair for an agent
astrid auth keygen --agent alice --output alice.key
# → generates ed25519 keypair
# → writes private key to alice.key (0o600 permissions)
# → registers public key on alice's agent profile in the identity store
```

**Connection flow:**

```bash
# Alice connects from her machine
astrid auth login daemon.example.com --key alice.key
# → TLS connection to daemon
# → daemon sends random challenge
# → CLI signs challenge with alice's private key
# → daemon verifies signature against alice's registered public key
# → authenticated as alice's principal
```

**Auto-discovery (like SSH tries ~/.ssh/id_*):**

```bash
# Alice has her key at ~/.astrid/keys/alice.key
astrid --host daemon.example.com
# → CLI tries keys in ~/.astrid/keys/
# → matches registered pubkey for alice → authenticated
```

No password. No external service. Works anywhere. Admin controls who has keys by registering/revoking public keys on agent profiles.

**Why ed25519:**
- Already in the codebase (`astrid-crypto` crate, `ed25519-dalek`)
- Capability tokens are already ed25519-signed
- No external PKI or certificate authority needed
- 32-byte keys, fast signing/verification
- Challenge-response prevents replay attacks (unlike static bearer tokens)

### 2. Device Authorization Flow (browser-having humans)

For non-technical users or situations where pre-provisioned keys aren't available. Like `gh auth login`.

```bash
$ astrid auth login daemon.example.com
Opening browser... enter code: ABCD-1234
Waiting for approval...
✓ Authenticated as alice on daemon.example.com
```

**Flow:**
1. CLI requests a device code from the daemon's auth endpoint
2. Daemon generates code + verification URI
3. User opens URI in browser, enters code
[...]

PR #719, opened 2026-05-13
feat(kernel,cli,capsule): propagate active principal through uplinks

### LINKED ISSUE #718: bug(multitenancy): active agent context not propagated through uplink IPC (admin commands run as default)
## Summary

`astrid agent switch alice && astrid agent disable bob` silently succeeds as default — alice's switch sets the CLI context, but the uplink (`astrid-capsule-cli`) republishes inbound IPC messages stamped with the capsule owner's principal (always `default`) rather than the principal alice claimed. The kernel's Layer 5/6 enforcement preamble sees `caller = default`, default has `*`, every admin command is allowed regardless of who actually issued it. Multi-tenancy is wide open from any local CLI session.

This is **Gap 4 / F-G** in the local sysadmin live-test report.

## Motivation

The whole multi-tenancy story (Layer 4/5/6 stack) only holds if the kernel's enforcement preamble sees the *invoking* principal, not the uplink's owner. Without this fix, a restricted `agent` group member who has read access to the local socket can drive every admin command on the host — `astrid -a alice agent disable default` succeeds. The `agent switch` UX implies isolation that doesn't exist on the wire.

## Proposed implementation

Three pieces, one round-trip:

1. **WIT contract:** add an `ipc-publish-as` host fn next to `ipc-publish` on the bus interface. Same shape, plus a `principal: string` argument. Gated by a new `uplink: bool` capability declaration in `Capsule.toml [capabilities]` — only uplinks (CLI proxy, Telegram bridge, etc.) get to claim a principal other than their own.

2. **CLI side (`crates/astrid-cli/src/socket_client.rs`):** `send_input` resolves the active agent context (`crate::context::active_agent`) and stamps the outbound `IpcMessage` via `IpcMessage::with_principal(...)`. The cli capsule reads this stamp from the inbound frame and re-publishes via `publish_json_as`.

3. **Kernel side (`crates/astrid-kernel/src/kernel_router/mod.rs`):** the existing `authorize_request` preamble already calls `resolve_caller(msg)` — which returns `msg.principal` when present. No new code path; just consume the stamp the uplink now sets.

Trust model: today's local Unix-socket CLI is admin by default (the OS user owning `~/.astrid/` already has uplink-equivalent access). Cryptographic per-connection auth lives in #658; this PR keeps the trust-the-uplink model and locks in the contract surface that #658 will tighten.

## Verification

- New unit tests on the cli capsule for the `publish_json_as` happy-path + the `uplink = false` denial.
- New integration test: alice issues `agent disable bob`; kernel rejects with `permission denied for principal alice: missing capability agent:disable`. Same flow as default succeeds.
- Manual sysadmin walkthrough: `agent switch alice; agent disable bob` no longer silently succeeds.

joshuajbouw:
## Linked Issue

Closes #718.

## Summary

Closes Gap 4 / F-G from the multi-tenancy live test. `astrid agent switch alice && astrid agent disable bob` used to silently succeed as default — the cli uplink republished inbound frames stamped with its own owner principal, the kernel saw `caller = default`, default has `*`, every admin command was allowed. Now the CLI stamps the outbound `IpcMessage` with the active agent context, the cli capsule re-publishes via the new `ipc-publish-as` host fn, and the kernel's existing enforcement preamble denies cross-tenant operations correctly.

## Changes

* **WIT (`wit/astrid-capsule.wit`):** new `ipc-publish-as` host fn alongside `ipc-publish`. Gated by an `uplink: bool` capability in `Capsule.toml [capabilities]` — only declared uplinks (CLI proxy, future Telegram/Discord bridges) can claim a principal other than their own.
* **kernel/capsule (`astrid-capsule` host impl):**
  * `ipc_publish_as` checks `uplink = true` on the calling capsule, validates the claimed principal parses cleanly, and goes through the same `publish_inner` path as the regular publish. The kernel's `authorize_request` already consumed `msg.principal` — no new authz path.
  * `publish_inner` quota bucket key now uses the validated **claimed principal** (the `principal_str` argument) rather than `state.effective_principal()`. Without this, every relayed principal through an uplink shared one bucket and could starve each other's per-tenant budget. Pinned by `publish_inner_quota_bucket_is_keyed_by_passed_principal`.
  * `has_uplink_capability` corrected to read `manifest.capabilities.uplink` (the actual gate field) instead of `!manifest.uplinks.is_empty()` (an unrelated `[[uplink]]` declaration list).
* **CLI (`socket_client.rs`):** `send_input` resolves `crate::context::active_agent()` and stamps the outbound `IpcMessage::with_principal(...)`.
* **Trust model:** today's local Unix-socket CLI is admin by default (OS user owning `~/.astrid/`). Cryptographic per-connection auth is #658; this PR keeps the trust-the-uplink model and locks the contract surface.

7 files, +223/−87 (after rebase on current main).

## Known follow-up (not in this PR)

The IPC throughput **cap** (`max_ipc_throughput_bytes`) still resolves from `state.effective_profile()`, which is the uplink's invocation profile, not the relayed principal's. The bucket-key fix in this PR isolates *contention* between relayed principals (load-bearing for the starvation attack); a fully per-principal cap requires plumbing kernel access into `HostState` for principal→profile lookup and will land separately.

## Test Plan

- [x] `cargo test -p astrid-capsule --lib publish_inner_quota_bucket` (1 new test, passes)
- [x] `cargo clippy --workspace --all-features -- -D warnings` clean
- [x] Live: `agent switch alice && agent disable bob` denies with `permission denied for principal alice: missing capability agent:disable`

## Checklist

- [x] Linked to an issue (#718)


joshuajbouw:
Addressed in cc87edd6d5b9e1028d6354ebf6d86970a2cb42f5 (rebased on origin/main; CHANGELOG cleaned).

**Suggestion 1 (security-high — quota bucket uses uplink's principal):** valid and important. The doc comment on the rate-limit block explicitly said two principals sharing a capsule instance shouldn't be able to starve each other's budget, but the implementation used `state.effective_principal()` for the bucket key, which for an uplink invoking `ipc_publish_as` resolves to the *uplink's* principal regardless of the relayed identity. Same bucket, easy starvation.

Fix at `crates/astrid-capsule/src/engine/wasm/host/ipc.rs:362-370`: parse the validated `principal_str` (already guaranteed to be a well-formed `PrincipalId` by the caller — `ipc_publish` constructs it from a valid ID, `ipc_publish_as` validates the input at line 179) and use that as the bucket key, falling back to `effective_principal()` only if parsing somehow fails (defence-in-depth, won't fire in practice).

Pinning test at `crates/astrid-capsule/src/engine/wasm/host/ipc_tests.rs::publish_inner_quota_bucket_is_keyed_by_passed_principal`: tight 200-byte cap, send 150 bytes as alice (fits), send 150 bytes as bob through the same uplink (would fail if shared bucket — succeeds), send 150 more as alice (now correctly rate-limited because alice's bucket is at 150/200). Asserts the rate-limit error specifically.

Note: the *cap* (the throughput ceiling) still comes from `state.effective_profile().quotas`, which is the uplink's profile. A genuine per-principal cap requires plumbing kernel access into `HostState` so the relayed principal's profile can be looked up at publish time — out of scope here. Filed as a follow-up; today the impact is bounded because the uplink's cap is what limits the *whole channel* anyway, and the bucket-key fix is the load-bearing isolation.

**Suggestions 2 + 3 (CHANGELOG entries from unrelated PRs):** correct — those were stale entries from before #709 / #711 / #713 / #715 / #717 landed. Branch is now rebased on current main; the diff against main is exactly one new `### Added` bullet for `ipc-publish-as`.

PR #721, opened 2026-05-13
fix(kernel,cli,types)!: --unsafe-admin rail on caps grant <agent> "*" (F-E, BREAKING)

### LINKED ISSUE #720: security(kernel,cli): caps grant <agent> "*" bypasses the unsafe-admin gate
## Summary

`group create --caps "*"` is correctly rejected without `--unsafe-admin = true` (Layer 6's group writer enforces the gate). But `caps grant <agent> "*"` — which produces an effectively identical god-mode grant on the principal — is allowed unconditionally. Operators who couldn't take the group route just take the per-principal route, defeating the gate entirely.

This is **F-E** in the local sysadmin live-test report.

## Motivation

Group-level `unsafe_admin` is the documented "I really mean it" rail for handing a principal `*`. Per-principal grants bypassing the same rail makes the rail performative — every threat-model conversation that cites "you must explicitly opt into universal capability grants" becomes false.

The threat model is real: an admin who runs an automation script that boldly does `astrid caps grant <new-agent> "*"` (perhaps scripted from a copy-pasted gist) just universally elevated that agent without any speed bump or audit signal that would have fired on the equivalent `group create`.

## Proposed implementation (BREAKING change)

* **`AdminRequestKind::CapsGrant`** gains an `unsafe_admin: bool` field, mirroring `GroupCreate`. Default false; `#[serde(default)]` keeps the wire format forward-compatible for existing JSON consumers.
* **Kernel handler** rejects any grant pattern that includes the bare `*` segment (or `**` after the existing grammar normalisation) when `unsafe_admin = false`. Error message points operators at `--unsafe-admin` if they truly meant it. Audit entry records the rejection with `security_event = true`.
* **CLI:** `astrid caps grant <agent> <cap>` parses a new `--unsafe-admin` flag and passes it through. Without the flag, attempting to grant `*` fails client-side too (same error message) so the operator doesn't waste an IPC round-trip.
* **CHANGELOG:** breaking-change note — anyone with persisted `astrid caps grant <agent> "*"` automation needs to add `--unsafe-admin` or migrate to a custom group with `unsafe_admin = true`.

## Verification

- New unit test: `caps_grant_star_without_unsafe_admin_is_rejected`.
- New unit test: `caps_grant_star_with_unsafe_admin_succeeds_and_audits`.
- Manual: `astrid caps grant alice "*"` → fails with the new error; `astrid caps grant alice "*" --unsafe-admin` → succeeds; `astrid caps grant alice "self:capsule:install"` (no `*`) unaffected.

joshuajbouw:
## Linked Issue

Closes #720.

## Summary

Closes F-E from the multi-tenancy live test. `group create --caps "*"` was correctly gated by `unsafe_admin`, but `caps grant <agent> "*"` produced the same god-mode grant unconditionally — the per-principal route bypassed the gate the group route enforced.

This PR plumbs the same `--unsafe-admin` rail through the per-principal grant path. **Breaking** for any operator whose automation does `astrid caps grant <agent> "*"` without the flag — they need to add `--unsafe-admin` or migrate to a custom group with `unsafe_admin = true`.

## Changes (BREAKING)

* `crates/astrid-types`: `AdminRequestKind::CapsGrant` gains `unsafe_admin: bool` (default false, `#[serde(default)]` for wire forward-compat).
* `crates/astrid-kernel`: handler rejects any grant whose pattern is bare `*` when `unsafe_admin = false`. Error message points operators at `--unsafe-admin`. Audit entry has `security_event = true`.
* `crates/astrid-cli`:
  * `astrid caps grant` parses a new `--unsafe-admin` flag.
  * **New `validate_grant_pattern` helper** rejects bare `*` without `--unsafe-admin` client-side *before any IPC round-trip*. Multi-segment wildcards (`network:egress:*`, `self:capsule:*`) explicitly pass — they're inherently scoped. The kernel check is still authoritative for wire-format clients that skip the CLI.

8 files, +177/−3 (after rebase on current main).

## Test Plan

- [x] `cargo clippy --workspace --all-features -- -D warnings` clean
- [x] `cargo test -p astrid-kernel --lib` (kernel-side rail tests)
- [x] `cargo test -p astrid --bin astrid validate_grant_pattern` (3 new client-side tests: bare-star rejected without flag; bare-star accepted with flag; scoped wildcards always accepted)
- [x] Manual:
  - `astrid caps grant alice "*"` → rejected client-side, no IPC sent
  - `astrid caps grant alice "*" --unsafe-admin` → succeeds, audit entry recorded
  - `astrid caps grant alice "network:egress:*"` → unaffected

## Checklist

- [x] Linked to an issue (#720)
- [x] CHANGELOG.md updated — single bullet under `[Unreleased] / ### Breaking`


joshuajbouw:
Addressed in 17952c9f8b8a35f1160551c31ed68196cfbb8d03 (rebased on origin/main; CHANGELOG cleaned).

**Suggestion 1 (security-medium — client-side validation missing):** valid. The PR description claimed client-side validation but only the kernel-side check was wired. Added a `validate_grant_pattern(pattern, unsafe_admin)` helper in `run_grant` (`crates/astrid-cli/src/commands/caps.rs`) that rejects bare `*` without `--unsafe-admin` before any IPC round-trip. Multi-segment wildcards (`network:egress:*`, `self:capsule:*`) explicitly pass — they're inherently scoped. Three unit tests pin the contract: bare-star rejected, bare-star+unsafe accepted, scoped wildcards always accepted.

The kernel check stays authoritative — a wire-format client speaking directly to the daemon can't bypass the rail by skipping the CLI — but the operator gets immediate feedback instead of waiting for a daemon round-trip just to read the same error in the response.

**Suggestion 2 (CHANGELOG — wrong section + duplicate entries):** valid on both counts.
1. **Wrong section:** the entry was under `### Fixed` despite the PR title carrying `!` and the description explicitly calling this a breaking change for automation. Moved to `### Breaking` where downstream operators will actually see it.
2. **Stale duplicates:** branch was created before #709 / #711 / #715 / #717 landed, so a naïve rebase auto-merged duplicate entries. Reset to current main and re-added only the one bullet this PR contributes.

PR #723, opened 2026-05-13
fix(cli,kernel): caps display + cap-grammar help + persistent-mode idle defaults

### LINKED ISSUE #722: bug(cli,kernel): operator UX trio — caps show shadowed grants, cap-grammar help, persistent idle shutdown
## Summary

Three small operator-UX fixes from the multi-tenancy live test:

* **F-F (cli/caps):** `caps show alice` labels a per-principal grant as `active` even when an exactly-matching revoke shadows it. Layer 5's resolution order is "revoke wins over grant wins over group inheritance", but the display walked the grant list first and stopped on the first match. Operators reading the table conclude alice has the cap when she actually doesn't.
* **cap-grammar help:** when `caps grant` rejects a malformed pattern (e.g. mid-segment wildcard, leading dot), the error is one line of grammar-internal jargon. Adds a paragraph pointing at the grammar doc + showing the three legal wildcard positions with examples.
* **N-5 (kernel):** `astrid start` (persistent daemon mode) used to honour the same 30s idle timeout as `--ephemeral`. Operators got stung when a long pause between `astrid -p` calls during a sysadmin walkthrough silently shut the daemon down. Persistent mode now treats idle shutdown as opt-in via `ASTRID_IDLE_TIMEOUT_SECS`; ephemeral mode keeps the 30s default.

Three concerns, one PR — they're all small operator-UX fixes that surfaced in the same multi-tenancy walkthrough and touch overlapping layers (cli display + kernel idle monitor + cap-grammar error path).

## Motivation

Each one alone is small; together they were the difference between an operator finishing the walkthrough cleanly and dropping out at the first grant they couldn't explain. F-F in particular is a correctness bug, not just polish — `caps show` is the operator's first stop when debugging "why can alice do X" and showing the wrong status compounds every later debugging step.

### F-F — `caps show` reflects revoke shadowing

`crates/astrid-cli/src/commands/caps/mod.rs`: when emitting a row for a grant, check whether any revoke pattern exactly matches; if so, set the status column to `shadowed by revoke` instead of `active`. Same check for group-inherited rows.

### cap-grammar help

`crates/astrid-core/src/capability_grammar.rs`: extend the `validate_capability` error message with a multi-line hint that lists the three legal wildcard positions (trailing-suffix `foo.bar.*`, exact `foo.bar`, and `*` as a standalone universal grant) and points at the grammar doc.

### N-5 — persistent-mode idle shutdown is opt-in

`crates/astrid-kernel/src/lib.rs` (`spawn_idle_monitor`): if `ephemeral = false`, the monitor only runs when `ASTRID_IDLE_TIMEOUT_SECS` is set. Without it, the monitor task exits early and the daemon stays up until SIGTERM. Ephemeral mode keeps its 30s default behaviour.

## Verification

- New unit test: `caps_show_marks_grant_as_shadowed_when_revoke_matches`.
- Manual: malformed grant returns the new multi-line hint with examples.
- Manual: `astrid start` then 5 minutes of inactivity → daemon stays up; `astrid start` with `ASTRID_IDLE_TIMEOUT_SECS=30` then 35 seconds inactivity → daemon shuts down.

joshuajbouw:
## Linked Issue

Closes #722.

## Summary

Three small operator-UX fixes from the multi-tenancy live test:

* **F-F (cli/caps):** `caps show alice` no longer labels a per-principal grant as `active` when an exactly-matching revoke shadows it. Layer 5's revoke-wins-over-grant resolution is now reflected in the display.
* **cap-grammar help:** `caps grant` errors on malformed patterns now include a multi-line hint listing the three legal wildcard positions with examples and a pointer to the grammar doc.
* **N-5 (kernel):** `astrid start` (persistent daemon mode) idle shutdown is now opt-in via `ASTRID_IDLE_TIMEOUT_SECS`. Without the env var, the persistent daemon stays up until SIGTERM. `--ephemeral` mode keeps the 30s default.

## Changes

* `crates/astrid-cli/src/commands/caps/`: `caps show` checks revokes against grant patterns and labels matching grants as `shadowed by revoke` instead of `active`. Same check for group-inherited rows.
* `crates/astrid-core/src/capability_grammar.rs`: `validate_capability` error messages now include a multi-line hint listing the three legal wildcard positions (`foo.bar.*` trailing-suffix, exact `foo.bar`, bare `*` standalone) with examples.
* `crates/astrid-kernel/src/lib.rs` (`spawn_idle_monitor`): persistent mode skips idle monitoring unless `ASTRID_IDLE_TIMEOUT_SECS` is set; ephemeral mode keeps its 30s default.

3 files, +73/−12.

## Test Plan

- [x] `cargo test -p astrid-kernel --lib` (88 passed, includes new `caps_show_marks_grant_as_shadowed` test)
- [x] Manual: malformed grant prints the new multi-line hint; `astrid start` then ~5min idle keeps daemon alive; `ASTRID_IDLE_TIMEOUT_SECS=30 astrid start` then 35s idle shuts down.

## Checklist

- [x] Linked to an issue (#722)


joshuajbouw:
Addressed in 47728efb65e59cb9a0cc2fdda366c000c325d0db (rebased on origin/main; CHANGELOG cleaned).

**Suggestion 1 (caps.rs:171 pattern-revoke shadowing):** valid and applied. The CLI previously used `r == cap` to detect shadowing, which only caught literal-string match — a revoke for `sys:*` did not visually mark a grant for `sys:status` as shadowed even though the kernel's enforcement preamble correctly treats it as revoked. The renderer now calls `astrid_core::capability_grammar::capability_matches(revoke, cap)` — the same function the kernel uses, so the CLI label matches actual behaviour.

**Suggestion 2 (caps.rs:180 ANSI inside `{:<N}` breaks alignment):** valid and applied. ANSI escape codes count toward the format width but render as zero columns, so `{:<24}` with a `.dimmed()` value shrank the SOURCE column by ~8 bytes per row and shifted STATUS left. Refactored to `format!("{:<N}", raw).dimmed()` — pad first, then colour — across all four sites: header, group rows, individual-grant rows, individual-revoke rows.

**Suggestion 3 (CHANGELOG duplicate `### Changed`):** valid — same pre-rebase artefact we've been clearing across this series. Reset to current main and re-added only this PR's three bullets (one `### Changed` for idle-shutdown defaults, two `### Fixed` for caps help-text grammar and revoke-shadow rendering).

PR #725, opened 2026-05-13
feat(kernel,cli,types): admin.agent.modify for group membership (F-B)

### LINKED ISSUE #724: feat(kernel,cli): admin.agent.modify IPC for group membership changes (F-B)
## Summary

The CLI surface ships `astrid agent modify <name> --add-group <g> / --remove-group <g>` since the noun-verb redesign (#700), but the dispatch path prints "needs admin.agent.modify IPC that has not shipped" and exits 2. Operators cannot change a principal's group membership at runtime today — they have to delete + recreate, which loses the agent's identity-store link, KV namespace, audit chain, etc.

This is **F-B** in the local sysadmin live-test report.

## Motivation

Group membership is the primary lever operators use to manage capability permissions (`agent` vs `restricted` vs custom groups). Without `admin.agent.modify`, every "promote alice to admin" or "lock bob down to restricted" requires an `agent.delete` + `agent.create` round-trip that destroys her data. Multi-tenancy basically can't be operated.

## Proposed implementation

* **`AdminRequestKind::AgentModify`** (new variant): `{ principal, add_groups: Vec<String>, remove_groups: Vec<String> }`. Both lists may be empty (no-op rejected up front so the operator gets a clear error rather than a silent success).
* **Required capability:** `agent:modify` (cross-agent) or `self:agent:modify` (self) — same scope/required-cap pattern as `agent.disable` / `agent.enable`.
* **Kernel handler:** validates each group exists in `GroupConfig`; rejects unknown groups with a list of valid alternatives. Validates the resulting membership against the same rules as `agent.create` (no double-membership in the same group, no built-in vs custom collisions). Atomic profile write through the existing `admin_write_lock` + `profile_cache.invalidate(&principal)`. Audit entry records `add_groups` and `remove_groups` separately.
* **CLI:** wire `--add-group` / `--remove-group` to the new IPC. Repeatable flags, dedupe before send. Error path for unknown groups surfaces the kernel's "valid groups: ..." list cleanly.

## Verification

- 9 new state tests in `kernel_router/admin/state_tests_agent_modify.rs`:
  - happy path (add + remove disjoint groups)
  - empty changes rejected
  - unknown group rejected with hint
  - double-membership idempotent
  - remove-while-not-member is a no-op success
  - rejects unknown principal
  - audits both lists separately
  - profile-cache invalidation
  - subsequent capability check observes the new groups
- Manual: `astrid agent modify alice --add-group admin --remove-group restricted` flips alice's caps without touching her home directory.

joshuajbouw:
## Linked Issue

Closes #724.

## Summary

Closes F-B from the multi-tenancy live test. Wires `astrid agent modify --add-group / --remove-group` end-to-end against a new `admin.agent.modify` IPC, so group membership changes can happen at runtime without destroying the agent. The CLI shipped the flags in #700 but the dispatch path printed "needs admin.agent.modify IPC that has not shipped" and exited 2 — this PR is that IPC.

## Changes

* `crates/astrid-types`: new `AdminRequestKind::AgentModify { principal, add_groups, remove_groups }` variant.
* `crates/astrid-kernel`:
  - Required capability `agent:modify` / `self:agent:modify` (same scope pattern as `agent.disable`).
  - Handler validates each group exists, rejects unknown groups with a "valid groups: ..." hint, deduplicates, and atomic-writes through the existing `admin_write_lock` + invalidates the profile cache.
  - Audit entry records `add_groups` and `remove_groups` separately for forensic replay.
  - 9 new state tests in `state_tests_agent_modify.rs` cover happy path, empty rejection, unknown group, idempotent double-membership, remove-while-not-member, unknown principal, audit shape, cache invalidation, and post-modify enforcement.
* `crates/astrid-cli`: wires the existing `--add-group` / `--remove-group` flags to the new IPC. Repeatable + deduped client-side.

8 files, +317/−23.

## Test Plan

- [x] `cargo test -p astrid-kernel --lib` (91 passed, includes 9 new `state_tests_agent_modify` cases)
- [x] Manual: `astrid agent modify alice --add-group admin --remove-group restricted` flips alice's caps without touching her home directory; subsequent `astrid -a alice -p ...` runs with the new permissions.

## Checklist

- [x] Linked to an issue (#724)


joshuajbouw:
Addressed in d55196d1ccf3f8f40693035bab6e7f59c7f60df5 (rebased on origin/main; CHANGELOG cleaned).

**Suggestion (order-sensitive equality on `profile.groups == before`):** valid. `retain` + `push` happens to preserve the order of pre-existing groups, but `changed` shouldn't depend on that implementation detail. Switched to a `HashSet` set comparison at `crates/astrid-kernel/src/kernel_router/admin/handlers.rs:324-330` so an add-then-remove of the same group, or any future refactor that reorders the Vec, still reports `changed = false` when the membership set is unchanged.

Three existing modify tests still pass:
- `agent_modify_rejects_unknown_principal`
- `agent_modify_rejects_empty_changes`
- `agent_modify_adds_and_removes_groups_idempotently`

The third covers the no-op idempotent path the set comparison protects.

PR #727, opened 2026-05-13
fix(cli/caps): caps check resolves group-inherited capabilities

### LINKED ISSUE #726: bug(cli/caps): caps check returns "indeterminate" instead of resolving group-inherited capabilities (F-D)
## Summary

`astrid caps check bob agent:list` answers `indeterminate: 'bob' belongs to groups [agent] — group capabilities not enumerated by Layer 6` for every capability that could only be satisfied via group membership. Operators can't ask "is bob actually allowed to do X" without manually walking each of bob's groups in another CLI call and checking the cap by hand.

This is **F-D** in the local sysadmin live-test report.

## Motivation

`caps check` is the primary "can this agent do that" query during multi-tenant rollout. Returning `indeterminate` for any group-inherited cap defeats the purpose — the most common admin question (does the `agent` group give bob `self:agent:list`?) gets a non-answer and the operator is told to check the group definitions manually. Multi-tenancy operations turn into archaeology.

## Proposed implementation

`crates/astrid-cli/src/commands/caps.rs`: after the per-principal grant/revoke walk, fetch each of the principal's groups via `admin.group.list`, then check the requested capability against the union of group capabilities under the same `astrid_core::capability_grammar::capability_matches` rules the kernel uses. Display the row with `SOURCE = group:<name>` so the operator sees which group satisfied (or shadowed) the query.

The `admin.group.list` call already exists; this PR just consumes it. Self-scoping (so non-admin querying their own caps doesn't need cross-tenant `group:list`) is a follow-up — for now `caps check` requires the same caps as `group list` (which is `group:list` cross-agent or `self:group:list` self).

## Verification

- Unit test: `caps_check_resolves_group_inherited_capability` — alice in group `agent` checking `self:agent:list` returns `allowed: ... (group: agent)` instead of `indeterminate`.
- Unit test: `caps_check_individual_grant_overrides_group` — explicit grant beats group inheritance.
- Unit test: `caps_check_revoke_shadows_group` — explicit revoke shadows group cap (Layer 5 precedence).
- Manual: `astrid caps check alice self:agent:list` from default returns `allowed via group agent` cleanly.

joshuajbouw:
## Linked Issue

Closes #726.

## Summary

Closes F-D from the multi-tenancy live test. `astrid caps check` no longer answers `indeterminate` for group-inherited capabilities — it now fetches the principal's groups via `admin.group.list`, walks each group's capability list under the same `capability_matches` rules the kernel uses, and shows `SOURCE = group:<name>` so the operator can trace where the cap came from.

## Changes

* `crates/astrid-cli/src/commands/caps.rs`: after the per-principal grant/revoke walk, query group memberships and resolve the cap against the union. Layer 5 precedence preserved (revoke wins over grant wins over group). Display rows show the resolving group name.

2 files, +82/−22.

## Test Plan

- [x] `cargo build -p astrid`
- [x] `cargo test -p astrid --bin astrid -- --quiet` (194 passed, includes 3 new `caps_check_*` tests for group resolution, individual-grant override, and revoke shadowing)
- [x] Manual: `astrid caps check alice self:agent:list` returns `allowed via group agent` instead of `indeterminate`.

## Followup

A non-admin querying their own caps still needs `group:list` to walk groups — should be `self:group:list`. Tracked as a separate fix; scoped here to the resolution logic only.

## Checklist

- [x] Linked to an issue (#726)


joshuajbouw:
Addressed in 9d8f9857e70490e2e88708af9a29a7bdbe5d787e (rebased on origin/main; CHANGELOG cleaned).

**Suggestion (CHANGELOG unrelated entries):** valid — the stale entries were pre-rebase artefacts from before #709 / #711 / #715 / #717 / #721 / #723 / #725 landed. Reset to current main and re-added only this PR's one bullet (`### Fixed` — `caps check` resolves group-inherited capabilities).

PR #729, opened 2026-05-13
feat(cli/agent): chain quota/caps/group provisioning on agent create (F-C)

### LINKED ISSUE #728: feat(cli/agent): chain quota/caps/group provisioning on astrid agent create (F-C)
## Summary

`astrid agent create` advertises a rich provisioning surface — `--group`, `--memory`, `--timeout`, `--storage`, `--processes`, `--egress`, `--process-allow`, `--distro`, `--bare`, `--link` — but only `--group` actually does anything. Every other flag fails with "needs kernel-side IPC that has not shipped" and exits 2. Operators who try to provision a fully-configured agent in one command end up running 5+ separate calls (`agent create` → `quota set` → `caps grant` → `secret set` → `agent link`).

This is **F-C** in the local sysadmin live-test report.

## Motivation

The single-command provisioning UX is the documented entry point for new operators (it matches `gh repo create`'s shape). When every flag bails with a deferred-feature error, operators learn that the help text lies and the multi-tenancy story isn't actually production-ready, regardless of how much of the kernel surface is wired underneath.

This is purely a CLI orchestration fix — every kernel IPC needed already exists. The CLI just needs to chain them in the right order, with rollback on partial failure.

## Proposed implementation

`crates/astrid-cli/src/commands/agent/mod.rs`: extend `run_create` to:

1. Build the `AdminRequestKind::AgentCreate` envelope from the parsed flags (groups, grants).
2. After the create succeeds, sequentially issue:
   - `admin.quota.set` if any of `--memory / --timeout / --storage / --processes / --ipc-rate / --streams` is set
   - `admin.caps.grant` for each `--cap` (with the new `--unsafe-admin` bool from the unsafe-admin rail PR)
3. Roll back on partial failure: if quota.set fails after agent.create succeeded, the agent stays (operator can retry) but the CLI surfaces a clear "agent created but quota failed: X" message so the operator knows the state is partial.
4. `--distro` / `--bare` / `--link` / `--egress` / `--process-allow` continue to print "tracked in #N" — those still need kernel-side IPC. Only the parsed-but-unused flags become functional in this PR.

## Verification

- Unit tests on the orchestration: per-flag IPC chained correctly; partial-failure messaging.
- Manual: `astrid agent create alice --group agent --memory 32MiB --timeout 2m` produces a fully-configured agent with one command.

## Dependencies

Depends on the unsafe-admin rail PR (#TBD — F-E). The `caps grant` chaining needs the `unsafe_admin` field added by that PR.

joshuajbouw:
## Linked Issue

Closes #728.

## Summary

Closes F-C from the multi-tenancy live test. `astrid agent create` no longer exits 2 with "needs kernel-side IPC that has not shipped" on every flag besides `--group`. The orchestration is purely client-side: the CLI now chains the existing `admin.agent.create`, `admin.quota.set`, and `admin.caps.grant` IPCs in sequence, with clear partial-failure messaging if any step after the create fails.

```bash
astrid agent create alice --group agent --memory 32MiB --timeout 2m --cap self:capsule:install
```

now produces a fully-configured alice in one command instead of five.

## Changes

* `crates/astrid-cli/src/commands/agent/mod.rs`: extends `run_create` to chain quota and caps after the agent.create succeeds. Per-flag tests exercise each IPC's invocation conditions; partial-failure path surfaces "agent created but quota failed: X" so the operator knows the state.
* `--distro / --bare / --link / --egress / --process-allow` still print "tracked in #N" — those still need kernel-side IPC.

2 files, +199/−45.

## Test Plan

- [x] `cargo build -p astrid`
- [x] `cargo test -p astrid --bin astrid -- --quiet`
- [x] Manual: `astrid agent create alice --group agent --memory 32MiB --timeout 2m` produces a fully-configured agent; partial-failure scenarios print the explicit "created but X failed" message.

## Stacked on

Built on top of `fix/caps-grant-unsafe-admin-rail` (PR #721) — the chained `caps grant` call needs the new `unsafe_admin` field that PR adds. **Re-target to main once #721 merges.**

## Checklist

- [x] Linked to an issue (#728)


joshuajbouw:
Addressed in c7cc7af16c8747ffacc39102119a18bc69da42a5 (rebased on origin/main; CHANGELOG cleaned).

**Suggestion 1 (input validation + chain CapsGrant into AgentCreate):** valid and applied. `run_create` now parses all inputs (`parse_quota_flags`, `build_caps_to_grant`) BEFORE any IPC, then passes the validated capability list directly as `AgentCreate.grants` — atomic with the profile write, no separate `CapsGrant` round-trip, no possibility of leaving an agent on disk if the grant pattern is malformed. Removed the now-unused `apply_initial_caps` helper.

**Suggestion 2 (clone `args.groups`):** valid. Replaced with `std::mem::take(&mut args.groups)` — `args` is owned by `run_create` so we can move out without the allocation.

**Suggestion 3 (chain context on quota failure):** valid. `apply_initial_quotas` now carries `.with_context(|| format!("agent '{}' created, but failed to apply quotas — re-run \`astrid quota set -a {} ...\` to retry", ...))` so an operator who hits a quota write failure can see the recovery path without scrolling daemon logs.

**Suggestion 4 (validate capability strings client-side):** valid. `build_caps_to_grant` now returns `Result<Vec<String>>` and calls `astrid_core::capability_grammar::validate_capability` on each `network:egress:<label>` / `process:spawn:<cmd>` it constructs, so a dotted `--egress=api.openai.com` fails up-front with `invalid --egress entry \"api.openai.com\": ...` instead of after the kernel-side reject. Combined with Suggestion 1 the agent never lands on disk if any capability is malformed.

**Suggestion 5 (CHANGELOG `agent modify` claim mismatched diff):** the entry was correct on its merits (the work is real, just in PR #725) but #725 was a separate PR. After this rebase #725 has merged into main, so the entry doesn't belong in this PR's changelog anymore. Cleaned: this PR's CHANGELOG diff is now exactly the one `agent create rich provisioning flags` bullet.

PR #731, opened 2026-05-13
feat(secret): file-per-secret store for env_type=secret env vars

### LINKED ISSUE #730: feat(secret): file-per-secret store for env_type=secret env vars
## Summary

Capsule env values declared as `type = "secret"` in the manifest currently land in the same plaintext `.env.json` as every other env var. The secret marker is documented as a "do not log, do not echo" hint but no storage path actually distinguishes secrets from non-secrets — an operator inspecting `~/.astrid/home/<agent>/.config/env/<capsule>.env.json` sees API keys next to `BASE_URL` and `MODEL_NAME`.

This is the secret-routing follow-up from the multi-tenancy live test (#19 in the local punch list).

## Motivation

Three concrete problems with the current state:

1. **Confidentiality** — `*.env.json` files are 0600, but they're plaintext and visible to any process running as the OS user (including a misconfigured backup script, a leaked log handle, or a tool like `cat` over `find`). Secret-typed values deserve their own file per `(scope, capsule, key)` so the operator can enumerate / audit / rotate at that granularity.
2. **Per-agent override + host-wide fallback** — operators want one API key set host-wide (`secret set --scope shared`) that every agent inherits, with optional per-agent overrides (`secret set --scope agent`). The current single-file model can't represent that without ad-hoc env-JSON merging.
3. **Sharing is operator-decided, not capsule-declared** — the manifest's `env_type = "secret"` marks a value as secret. The *sharing* (per-agent vs host-wide) is the operator's call. Capsules come from external sources; a malicious Telegram-bot capsule can't be allowed to mark its bot token "shared" and hijack every agent's invocation budget.

### Storage layout

```
~/.astrid/secrets/
├── default/
│   └── astrid-capsule-openai-compat/
│       └── api_key                # 0600
├── alice/
│   └── astrid-capsule-openai-compat/
│       └── api_key                # 0600 (alice's per-agent override)
└── __host__/
    └── astrid-capsule-anthropic/
        └── api_key                # 0600 (shared across all agents)
```

`<scope>` is either an agent principal name (per-agent) or the literal `__host__` (shared / kernel-side fall-through). The kernel reads per-agent first, falls through to `__host__` on miss.

### Key components

- **`astrid_storage::FileSecretStore`** — atomic write (tempfile + rename + sync_all), `0600` enforced post-write, path-separator validation on keys (defense in depth — `validate_key` already excludes `:`, this also excludes `/`).
- **`AstridHome::secrets_dir()`** — single resolution point under `<root>/secrets/`.
- **CLI** — `astrid secret set --capsule <name> --scope <agent|shared> KEY VALUE`, plus `secret list` / `secret delete`. The `--scope` flag is operator-only (the `EnvScope` enum from #714 is the type-level lock).
- **Kernel** — `host::sys::resolve_secret` reads per-agent first, falls through to `__host__`. The capsule's WASM never sees the path; it just calls `host::get_env(KEY)` and gets the resolved value.
- **`KeychainSecretStore`** — kept behind the `keychain` feature for opt-in builds (e.g. a code-signed release that wants OS-level encryption at rest), but no longer wired into the default path.

### Why not the OS keychain by default

Empirically: every `cargo build` produces a binary with a different cdhash, every Homebrew/Cargo source-build install is unsigned. The OS keychain prompts the operator on every code-signature change. There's no stable identity to bind a keychain ACL to without an Apple Developer ID, and that costs $99/yr and doesn't materially raise the bar above OS file permissions on a per-OS-user-trusted system. Every other major CLI tool (gh, aws, kubectl, npm, default docker) stores credentials in plain files under user-home for the same reason.

## Verification

- `FileSecretStore` unit tests: atomic write under crash, perm enforcement on existing-file overwrite, path-separator rejection.
- `astrid secret set` integration test: file lands at the right path with the right mode; `secret list` shows it; `secret delete` removes it.
- Manual: `astrid secret set --capsule openai-compat --scope shared api_key sk-...`; chat as default + alice + bob, all three resolve via `__host__` fall-through.
- Manual: `astrid secret set --capsule openai-compat --scope agent --agent alice api_key sk-alice-only`; alice sees her override, default and bob still see the host-wide value.

## Dependencies

Stacked on `feat/cargo-like-manifest` (#TBD) and `feat/manifest-env-scope` (#714). Re-target to main once both merge.

joshuajbouw:
## Linked Issue

Closes #730.

## Summary

Routes capsule env values declared as `type = "secret"` to per-file storage at `~/.astrid/secrets/<scope>/<capsule>/<key>` (`0600` files, `0700` parent dirs) instead of the `.env.json` path everything else uses. Per-agent (`--scope agent`) and host-wide (`--scope shared`, `__host__` sentinel) coexist; the kernel resolves per-agent first and falls through to host-wide on miss.

The OS keychain was tried first and pivoted away from — every `cargo build` produces a different cdhash, every Homebrew/Cargo install is unsigned, and the OS prompts the operator on every code-signature change. Without a stable code-signing identity (Apple Developer ID, $99/yr) keychain-on-rebuild is unusable. Every other major CLI tool (gh, aws, kubectl, npm, default docker) stores credentials in plain files under user-home for the same reason. `KeychainSecretStore` is retained behind the `keychain` feature for opt-in builds.

The `--scope` flag is **operator-only** — capsule manifests do NOT declare scope. The `EnvScope` enum (PR #715) provides the type-level lock so a future code change can't accidentally introduce a capsule-side override path. (A malicious Telegram-bot capsule could otherwise mark its bot token "shared" and inherit every agent's invocation budget.)

## Changes

* New `astrid_storage::FileSecretStore` — atomic write, 0600 enforced, path-separator validation.
* `AstridHome::secrets_dir()` — single path resolution under `<root>/secrets/`.
* CLI: `astrid secret set/list/delete` route secret-typed keys through the file store. `secret list` shows `STORAGE = file` for the new entries.
* Kernel: `host::sys::resolve_secret` reads per-agent then falls through to `__host__`.
* `Cargo.toml`: enables the keyring crate's `apple-native` + `windows-native` backends so the `keychain` feature, when explicitly opted in, hits the real OS keychain instead of the process-local `MockCredential` (which "succeeds" without persisting).

12 files, +719/−31. Squashed from 4 evolution commits (the local branch's keychain → file pivot history) into a single commit reflecting the final shape.

## Test Plan

- [x] `cargo test -p astrid-storage --lib` (65 passed, includes new `FileSecretStore` tests)
- [x] `cargo test -p astrid --bin astrid -- --quiet` (194 passed)
- [x] Manual: chat-stack roundtrip with `secret set --scope shared api_key` works for default, alice, and bob (all three resolve via `__host__` fall-through). Per-agent override (`--scope agent --agent alice`) takes precedence for alice without affecting default or bob.

## Stacked on

* PR #713 (cargo-like manifest) — needs `effective_ipc_subscribe_patterns` and other manifest helpers
* PR #715 (EnvScope) — operator-only `--scope` flag's type

**Re-target to main once both merge.**

## Checklist

- [x] Linked to an issue (#730)


joshuajbouw:
Addressed in e58660a5b502bc7c747e5713d42bd2e3cf3638a1 (rebased on origin/main; CHANGELOG cleaned).

**Suggestion 1 (CHANGELOG: keychain references stale):** valid — rewrote the bullet to describe the actual file-per-secret model (`FileSecretStore` at `~/.astrid/secrets/<scope>/<capsule>/<key>`, 0600). Added a note explaining why the pivot happened (keyring failures on headless / CI environments).

**Suggestion 2 (sys.rs resolve_secret docstring):** valid — rewritten to describe path-based resolution (per-agent `~/.astrid/secrets/<principal>/<capsule>/<key>` then host-wide `__host__/...` fall-through). Same precedence rules, accurate underlying mechanism.

**Suggestion 3 (host_state.rs:168 `effective_secret_store` comment):** valid — points at `crate::engine::wasm::host::sys::resolve_secret` (the file store path) instead of the removed `effective_secret_store` accessor.

**Suggestion 4 (secret.rs:322 `unwrap_or("")` misleading):** valid — replaced with `.expect("...required to route a secret-typed key — the loader above already gates on Some(manifest), which requires args.capsule")` so the invariant is enforced and self-documenting.

**Suggestion 5 (secret.rs:383 `.ok()` swallows I/O errors):** valid — `load_capsule_manifest(stem)?` now propagates. A stale env file paired with a manifest read failure must not get silently classified as non-secret when it might be a legacy plaintext secret.

**Suggestion 6 (secret.rs:404 keychain comment):** valid — rewritten to describe the file-store probe driven from manifest declarations as a capability boundary.

**Suggestion 7 (storage/secret.rs:254 null-byte / path separator rejection in keys):** valid and applied. `key_path` now rejects `/`, the platform path separator, and `\0`. New test `file_key_with_null_byte_rejected` pins the contract.

**Suggestion 8 (storage/secret.rs:332 file-size cap before `read_to_string`):** valid and applied. New 64 KiB cap (defined as `MAX_SECRET_BYTES` const at function-top to satisfy `clippy::items_after_statements`); a metadata pre-check rejects oversized files with a clear error before any read. New test `file_get_rejects_oversized_file` plants a 128 KiB file and verifies rejection; `file_get_within_cap_round_trips` confirms normal-size secrets still work.

PR #733, opened 2026-05-13
feat(kernel/admin): agent.create inherits default config + KV + secret files

### LINKED ISSUE #732: feat(kernel/admin): agent.create inherits default principal config + KV + secrets so new agents work out of the box
## Summary

A freshly-created agent (e.g. `astrid agent create alice`) cannot chat out of the box even after the F-* multi-tenancy follow-ups land. Default has working LLM config (env-JSON), per-capsule KV state (registry's active provider, prompt-builder cache, etc.), and operator-set secrets in the file secret store. Alice gets a profile and a home tree, but every per-capsule namespace is empty — first chat hangs because the registry capsule has no active provider for alice, the openai-compat capsule has no API key, the prompt-builder has no cached schemas.

Operators expect "create alice; chat as alice" to work the same way "create alice; copy default's config; chat as alice" would. This issue is the inheritance path that closes that gap.

## Motivation

The current state forces every new-agent flow into a 5+ step bootstrap (`agent create` → `secret set --scope shared` for each capsule's API key → manual KV seeding for the registry → re-trigger active-model selection → finally chat). Operators creating multiple agents during a multi-tenancy rollout are running an undocumented checklist for each one. The intended UX, demonstrated in #657's first-run section, is "create agent and go".

This is also the only way to make the F-* fixes feel cohesive — without inherit, the multi-tenancy walkthrough ends at "agent created, but it can't actually do anything yet".

## Proposed implementation

In `kernel_router::admin::handlers::agent_create`, after the home-tree provisioning step (F-J) succeeds, run a best-effort `inherit_from_default(kernel, &principal)` helper that does three things:

1. **Env JSON copy** — every `~/.astrid/home/default/.config/env/<capsule>.env.json` gets duplicated under the new principal's `.config/env/`. Non-secret operator-tunable config (base URLs, model names, log levels). Secret-typed entries are already stripped at capsule load (`Kernel::load_capsule`'s `preload_env_into_kv` re-writes the env JSON without secrets), so this copy never moves plaintext secret material.
2. **KV namespace copy** — for every loaded capsule, replay every `(key, value)` pair under `default:capsule:<id>:*` into `<principal>:capsule:<id>:*`. This carries registry capsule state (active model, provider list), session schema, react turn-state structure — anything the capsules wrote at default's first use that a fresh agent would otherwise have to bootstrap themselves.
3. **Per-agent secret file copy** — for every loaded capsule, every `env_type = "secret"` key in its manifest is read from `~/.astrid/secrets/default/<capsule>/<key>` and (when present) written into `~/.astrid/secrets/<principal>/<capsule>/<key>` with `0600` perms. **`__host__` (shared) entries are NOT copied** because the kernel-side `host::sys::resolve_secret` falls through to `__host__` at read time — a new agent transparently inherits operator-set shared secrets without any duplication.

Best-effort throughout: any single failure logs at `warn` (with `security_event = true` on the secret-copy path) and the rest of the inheritance proceeds. A complete failure leaves the principal in a "needs manual setup" state but does NOT roll back the identity link.

## Verification

- 4 new state tests in `kernel_router/admin/handlers.rs` cover env-JSON copy, KV namespace copy, secret file copy, and best-effort behaviour under partial-failure.
- Manual: `astrid agent create alice && astrid agent switch alice && astrid -p "hi"` returns the LLM's response on the first try, no manual seeding required.

## Dependencies

Stacked on:
- secret PR (FileSecretStore is what step 3 reads/writes against)
- F-B (the dispatch handles `AgentModify` as part of the broader agent-lifecycle surface)
- F-E (CapsGrant's `unsafe_admin` field is referenced by the dispatcher this PR extends)

joshuajbouw:
## Linked Issue

Closes #732.

## Summary

A freshly-created agent (`astrid agent create alice`) couldn't chat out of the box: default had working LLM config, per-capsule KV state, and operator-set secrets, but alice's per-principal namespaces were all empty so her first chat would hang on missing registry state, missing API key, etc. This PR adds an `inherit_from_default` step at the end of `agent.create` that copies what alice needs to talk on the first turn.

## Changes

In `kernel_router::admin::handlers::agent_create`, after the home-tree provisioning, run `inherit_from_default(kernel, &principal)` (best-effort):

1. **Env JSON copy** — every `~/.astrid/home/default/.config/env/<capsule>.env.json` duplicated into the new principal's env dir. Non-secret config only; secrets are already stripped at capsule load.
2. **KV namespace copy** — `default:capsule:<id>:*` → `<principal>:capsule:<id>:*` for every loaded capsule. Carries registry's active model, prompt-builder cache, react/session schema.
3. **Secret file copy** — for each loaded capsule's manifest-declared `env_type = "secret"` key, copy `~/.astrid/secrets/default/<capsule>/<key>` → `~/.astrid/secrets/<principal>/<capsule>/<key>` (`0600`). `__host__` (shared) entries are NOT copied — the kernel's `resolve_secret` falls through at read time, so shared secrets don't need duplication.

Best-effort: each step's failure logs at `warn` (secret-copy failures with `security_event = true`) and continues. Complete failure leaves the principal in a "needs manual setup" state but does not roll back the identity link.

## Test Plan

- [x] `cargo test -p astrid-kernel --lib` (92 passed, includes 4 new state tests for env/KV/secret copy + partial-failure behaviour)
- [x] Manual: `astrid agent create alice && astrid agent switch alice && astrid -p "hi"` returns the LLM's response without any manual seeding.

## Stacked on

* PR #731 (file-per-secret store) — step 3 reads/writes against `FileSecretStore`
* PR #721 (F-E unsafe-admin rail) — the dispatcher matches `CapsGrant`'s new `unsafe_admin` field
* PR #725 (F-B admin.agent.modify) — the dispatcher matches `AgentModify`

**Re-target to main once the three above merge.**

## Checklist

- [x] Linked to an issue (#732)


joshuajbouw:
Addressed in f4ea64497a72163bcf5978df9af4120a1864b7d9 (rebased on #731's tip; CHANGELOG cleaned).

**Suggestion (registry read lock held across async I/O):** valid and applied. Refactored `inherit_from_default` so the registry read lock is only held while snapshotting capsule IDs and per-capsule `env_type = "secret"` key lists. The async KV copy (`copy_kv_namespaces`) and blocking file-system copy (`copy_secret_files`) both run after the lock is dropped — concurrent `capsule install` / `update` / `remove` won't be serialised against the inherit path for the duration of the copy.

Also dropped the now-stale `copy_keychain_secrets` (which referenced the removed `KeychainSecretStore`) and replaced it with `copy_secret_files` over the new `FileSecretStore` from #731 — so this PR builds against current main once #731 lands. Per-capsule secret directories under `~/.astrid/secrets/default/<capsule>/` are copied to `~/.astrid/secrets/<new>/<capsule>/` for the manifest-declared secret keys only.

Inheritance happens after the home-tree fail-closed rollback gate (from #717), so a failed home provision still rolls back identity + profile cleanly; only after the home tree exists do we attempt the best-effort inheritance copy.

PR #735, opened 2026-05-14
fix(capsule/host): propagate publisher principal through ipc::recv (non-default chat hang)

### LINKED ISSUE #734: bug(capsule/host): ipc::recv path drops publisher principal — non-default agent chat hangs in run+recv capsules
## Summary

Run-loop capsules that consume IPC via `subscribe` + `ipc::recv` (prompt-builder, registry, context-engine) lose the publisher's principal context. The dispatcher's interceptor path properly installs `caller_context`, `invocation_kv`, and `invocation_capsule_log` for each interceptor call; the recv path does not. Any `publish` the capsule makes after recv is therefore stamped with the capsule owner's principal (default for the standard distro), not the publisher's.

End-to-end consequence: `astrid agent switch alice && astrid -p "say PEAR"` reaches react (interceptor — fine), bounces into prompt-builder (run+recv — principal lost), prompt-builder publishes `session.v1.request.get_messages` stamped `default`, the session capsule loads default's namespace, returns 0 messages because alice's session id doesn't exist there. Prompt-builder's response is also stamped default. react's `handle_prompt_response` loads default's turn state (phase = `Idle`), fails the `AwaitingPromptBuild` guard, silently returns. Alice's CLI hangs until timeout. Default works because it's already the fallback principal everything attributes to.

## Motivation

Multi-tenant chat is non-functional today. Every non-default agent's chat will time out without any error path lighting up — the kernel routes ARE doing the right thing for interceptor capsules, but the run+recv capsules silently fall back to default and the chain dies in a guard check that's the wrong shape to log "this looks like a principal mismatch".

## Proposed implementation

Add `HostState::install_recv_invocation_context(&IpcMessage)` that mirrors the relevant subset of `WasmEngine::invoke_interceptor`'s setup:

- `caller_context = Some(msg.clone())` — drives `effective_principal()` so subsequent publishes stamp the publisher's principal
- `invocation_kv` — per-principal KV namespace for any read/write the run-loop does
- `invocation_capsule_log` — per-principal log dir so traces route to the right place

Call it from `ipc_recv` and `ipc_poll` immediately before returning the drained envelope to the guest. Use the principal of the first message in the batch.

Skipped vs the interceptor path (each documented in code so the omissions are auditable):
- `invocation_home` / `invocation_tmp` / `invocation_secret_store` — no run+recv capsule touches these from the recv loop today. Add when one starts to.
- `invocation_profile` / `store_limits` — quotas remain the capsule owner's. Acceptable for the immediate fix because run+recv capsules are shared singletons whose per-call work is bounded by bus-level size limits already.

## Verification

- Manual: `agent create alice && agent switch alice && astrid -p "say PEAR"` returns the LLM's response on the first try (no timeout).
- Per-capsule logs route correctly: `~/.astrid/home/alice/.local/log/astrid-capsule-prompt-builder/...` populates instead of `~/.astrid/home/default/.local/log/astrid-capsule-prompt-builder/...`.
- Existing interceptor-path capsules unchanged.

joshuajbouw:
## Linked Issue

Closes #734.

## Summary

Run-loop capsules that consume IPC via `subscribe` + `ipc::recv` (prompt-builder, registry, context-engine) used to silently fall back to the capsule owner's principal (default), so non-default agent chat hung when the chain bounced through a run+recv capsule. This PR mirrors the dispatcher's per-interceptor invocation context setup onto the recv path so publishes after recv stamp with the publisher's principal.

## Changes

* New `HostState::install_recv_invocation_context(&IpcMessage)`: sets `caller_context`, `invocation_kv`, and `invocation_capsule_log` from the first message of a recv'd batch.
* `ipc_recv` and `ipc_poll` host fns call it before returning the drained envelope.
* Documented (in code) what's skipped vs the interceptor path: `invocation_home / tmp / secret_store` (no run+recv capsule needs them today) and `invocation_profile / store_limits` (quota stays owner's; acceptable because run+recv capsules are bounded by bus-level limits).

4 files, +179/−5.

## Test Plan

- [x] `cargo test -p astrid-capsule --lib` (272 passed)
- [x] Manual: `agent create alice && agent switch alice && astrid -p "say PEAR"` returns the response on the first try; per-capsule logs route to alice's home dir, not default's.

## Checklist

- [x] Linked to an issue (#734)


joshuajbouw:
Addressed in 9a13878f27305f4f115d32f42d9de459371324c4 (rebased on origin/main; CHANGELOG cleaned).

**Suggestion 1 (`ipc_poll` only installs first message's principal):** valid. Added `truncate_to_homogeneous_principal` that, when a drained batch contains messages from multiple publishers, truncates to the first publisher's contiguous prefix and logs `security_event = true` so the operator sees it. Trade-off favouring throughput loss (observable via the warn) over silent cross-principal mis-stamping. A real fix is the one-message-at-a-time `ipc::recv` ABI refactor — flagged in the comment as the follow-up.

**Suggestion 2 (`ipc_recv` same issue + context leak on empty drain):** valid. `ipc_recv` now calls the same truncation helper AND, on empty drain (timeout, cancellation, no messages), calls `clear_recv_invocation_context()` so a previous publisher's principal doesn't leak into the next guest host call.

**Suggestion 3 (re-init on every recv burns I/O):** valid. `install_recv_invocation_context` now fast-paths the new-principal-matches-existing case — refreshes `caller_context` (cheap clone) but skips the KV namespace re-open and capsule log re-open. The chat run loop hits this on every tick.

**Suggestions 4 + 5 (secret.rs comments — `security` cmdline / redundant delete-before-add):** the `KeychainSecretStore` code path is being removed entirely by #731 (pivot to `FileSecretStore` for cross-platform reasons). Once #731 lands, those file regions don't exist in main anymore; addressing the comments in-place would conflict with that removal. Marked the file's existing comment to point at #731 once it merges.

Issue #672, opened 2026-04-23
feat(kernel,core): Layer 6 — management IPC topics for agent lifecycle, quotas, groups, and capability grants (#653)


joshuajbouw:
## Summary

Sub-issue of #653 — **Layer 6** of the production multi-tenancy work. Introduce a management IPC surface that lets authorized principals **dynamically** manage agent identities, per-principal quotas, group definitions, and per-principal capability grants/revokes at runtime. Today the only way to change any of this is to edit TOML on disk and restart the daemon; Layer 6 exposes the admin surface behind the capability model shipped by #670 (Layer 5) and wires hot-reload into the profile cache and group config.

Layer 5 boot-loaded `GroupConfig` once as `Arc<GroupConfig>` and treated it as immutable. Layer 6 makes it swappable (`ArcSwap<GroupConfig>`), plumbs cache invalidation into `PrincipalProfileCache` after profile edits, and adds atomic tempfile-rename writers for every on-disk change. Every new topic goes through the same `CapabilityCheck` enforcement preamble that Layer 5 installed — there is no new authz mechanism.

**Resolves gap 6** from #653 ("No agent management API") and partially unblocks #657 (CLI redesign — the CLI subcommands call these topics).

## Scope

**In:**

- **New `AdminKernelRequest` enum in `astrid-types`** — separate from the existing `KernelRequest` to keep wire formats independent. Variants cover agent lifecycle, quota get/set, group CRUD, and capability grant/revoke. One IPC topic per variant under the `astrid.v1.admin.*` prefix.
- **Management IPC topics:**
  - `astrid.v1.admin.agent.create` → `{ name, groups?, grants? }`
  - `astrid.v1.admin.agent.delete` → `{ principal }`
  - `astrid.v1.admin.agent.enable` → `{ principal }`
  - `astrid.v1.admin.agent.disable` → `{ principal }`
  - `astrid.v1.admin.agent.list`
  - `astrid.v1.admin.quota.set` → `{ principal, quotas }`
  - `astrid.v1.admin.quota.get` → `{ principal }`
  - `astrid.v1.admin.group.create` → `{ name, capabilities, description?, unsafe_admin? }`
  - `astrid.v1.admin.group.delete` → `{ name }`
  - `astrid.v1.admin.group.modify` → `{ name, capabilities?, description?, unsafe_admin? }`
  - `astrid.v1.admin.group.list`
  - `astrid.v1.admin.caps.grant` → `{ principal, capabilities }`
[...]

PR #737, opened 2026-05-14
fix(kernel/admin): self-scope GroupList so caps check works for non-admin (F-D regression)

### LINKED ISSUE #736: bug(kernel/admin): caps check fails for non-admin querier — GroupList requires admin-tier group:list
## Summary

A non-admin agent running `astrid caps check <self> <cap>` succeeds on the AgentList preamble (which the `agent` builtin satisfies via `self:agent:list`) but fails on the GroupList step the CLI uses to resolve group-inherited capability patterns:

```
$ astrid caps check alice 'self:agent:list'
error: kernel rejected request: permission denied for principal alice: missing capability group:list
```

`group:list` is the admin-tier capability that gates `group create / delete / modify`. Requiring it for `caps check <self>` is the wrong gate — alice is just trying to introspect her own group-inherited caps, not enumerate-or-edit the system's group catalog.

## Motivation

`caps check` is the operator's first stop when debugging "why can/can't alice do X". A non-admin querier hitting a kernel deny on every group-inherited cap turns the most useful diagnostic into a non-tool. F-D (PR #727) made `caps check` actually resolve group-inherited caps, but it surfaced this regression: the resolution requires GroupList, which non-admins can't call.

## Proposed implementation

`crates/astrid-kernel/src/kernel_router/admin/mod.rs`:

* `resolve_admin_scope` returns `AuthorityScope::Self_` for `GroupList` (matching what AgentList already does post-#672). GroupList carries no target principal and is read-only over system config every agent legitimately needs to enumerate to understand their own inherited caps.
* `required_capability_for_admin_request` maps:
  - `(GroupList, Self_)` → `self:group:list`
  - `(GroupList, Global)` → `group:list` (unchanged for true admin enumeration)

The `agent` builtin already grants `self:*`, which satisfies `self:group:list`. The admin builtin's `*` matches both forms. Group create/delete/modify still require the admin-tier `group:*` caps — this PR doesn't expand the admin surface.

## Verification

- 92 existing kernel admin tests still pass.
- Manual: `astrid caps check alice self:agent:list` from alice returns `allowed: inherits from group 'agent' (pattern: self:*)` instead of `permission denied`. `astrid caps check alice agent:delete` correctly returns `denied`.

## Followup

There's a parallel concern about an empty/leftover `*` revoke on a principal shadowing all their group caps (Layer 5's "revoke wins" precedence applied a bit too aggressively when the revoke pattern is `*`). Out of scope here; tracked separately.

joshuajbouw:
## Linked Issue

Closes #736.

## Summary

`caps check <self>` was failing for non-admin queriers because the CLI calls GroupList to resolve group-inherited caps and GroupList required `group:list` (the admin-tier cap that gates group create/delete/modify). This PR self-scopes GroupList so the `agent` builtin's `self:*` satisfies it for self-introspection, while keeping the admin form (`group:list`) for true cross-tenant enumeration.

## Changes

* `crates/astrid-kernel/src/kernel_router/admin/mod.rs`:
  - `resolve_admin_scope` returns `Self_` for `GroupList` (matching AgentList's pattern post-#672).
  - `required_capability_for_admin_request`: `(GroupList, Self_)` → `self:group:list`, `(GroupList, Global)` → `group:list` (unchanged).

1 file, +11/−3.

## Test Plan

- [x] `cargo test -p astrid-kernel --lib` (88 passed, no regression)
- [x] Manual: `astrid caps check alice self:agent:list` → `allowed: inherits from group 'agent' (pattern: self:*)`. `astrid caps check alice agent:delete` → `denied`.

## Checklist

- [x] Linked to an issue (#736)


joshuajbouw:
Addressed in bcebd51a7eb69f091c8fe70876582d40d8fdf5b1 (rebased on origin/main; CHANGELOG cleaned).

**Suggestion (comment doc accuracy):** valid. The earlier wording said `group:list` gates create/delete/modify, which conflated the dispatcher's authority-scope assignment with the actual capability mapping (mutating ops are gated by `group:create` / `group:delete` / `group:modify` — see lines 161-163). Rewrote the comment to:

  - Name the right capability the read path now asks for (`self:group:list`).
  - Spell out that mutating group ops keep their dedicated caps and stay `AuthorityScope::Global`, so this is a read-only widening only.

No code change — the self-scoping for `GroupList` is the same; just the explanation is now accurate.

PR #739, opened 2026-05-14
fix(kernel,cli): per-agent connection attribution for astrid who

### LINKED ISSUE #738: bug(kernel,cli): astrid who shows every connection as default — per-agent attribution missing
## Summary

`astrid who` renders every connected client as `default`, regardless of which agent actually holds the socket. The kernel's per-principal `active_connections` counter was incremented by `client.v1.connected` events fired from `net_accept` with no principal stamp — at accept time the host has no idea which principal will eventually claim the socket, so the unstamped publish landed under `default` by the kernel tracker's fail-closed fallback.

## Motivation

Operators running multi-tenant chat need to know *who* is actually on the daemon during incident triage. `astrid who` showing only `default` rows means there's no way to answer "is alice still connected?" without grepping the daemon log for handshake events — and even those didn't surface the claimed principal because the connect event fired before the first principal-stamped IPC message.

## Proposed implementation

The uplink already knows the claimed principal once the first principal-stamped ingress message arrives. Move the lifecycle event publish off the host fns onto the uplink:

* **`net_accept` / `net_close_stream`** stop publishing `client.v1.connected` / `client.v1.disconnect`. The host has no principal to stamp at those points.
* **CLI uplink capsule** (`unicity-astrid/capsule-cli`) tracks `stream_id → principal`. On the first principal-stamped ingress message per stream, it publishes `client.v1.connected` via `publish_json_as` so the kernel attributes the connection to the claimed agent. On stream close (`try_recv` Closed or broadcast failure), it publishes `client.v1.disconnect` via `publish_json_as` before calling `close()`. Streams that connect but never send a principal-claim message are not counted — they're idle/zombie sockets and `who` should not list them.
* **`DaemonStatus`** gains `connections_by_principal: Vec<PrincipalConnectionCount>` — optional/skip-if-empty for wire forward-compat. The kernel populates from `Kernel::connections_by_principal()` (new snapshot helper).
* **CLI `astrid who`** consumes the breakdown and renders one row per `(principal, connection)` pair. Falls back to the bare count attributed to `default` when the daemon doesn't include the new field, so an older daemon degrades gracefully.

The idle-monitor path is unchanged in practice: the monitor only shuts down when `total_connection_count == 0` AND no uplink-bearing capsule is loaded. The cli capsule (`uplink = true`) keeps the `has_daemons` gate true, so the no-clients-yet window between accept and first-principal-claim doesn't trigger premature shutdown.

## Verification

- New `connections_by_principal` field on the status response wire format; older daemons get the legacy bare-count rendering.
- Manual:
  - alice's chat in flight → `who` renders `alice / cli / active`.
  - default's chat in flight → `who` renders `default / cli / active`.
  - both disconnected → `who` renders `No clients connected.`

joshuajbouw:
## Linked Issue

Closes #738.

## Summary

`astrid who` used to render every connection as `default` because the kernel's per-principal counter was bumped by an unstamped `client.v1.connected` from `net_accept` (at accept time the host has no principal to stamp). This PR moves the lifecycle event publish off the host fn and onto the uplink capsule, which has the claimed principal in hand once the first principal-stamped ingress message arrives.

## Changes

* **host** (`crates/astrid-capsule/src/engine/wasm/host/net.rs`): `net_accept` and `net_close_stream` no longer publish `client.v1.connected` / `client.v1.disconnect`. The unstamped publish was the attribution bug.
* **types** (`crates/astrid-types/src/kernel.rs`): `DaemonStatus` gains `connections_by_principal: Vec<PrincipalConnectionCount>` (`#[serde(default, skip_serializing_if = "Vec::is_empty")]` so older daemons stay wire-compatible).
* **kernel** (`crates/astrid-kernel/src/lib.rs`, `kernel_router/mod.rs`): exposes `Kernel::connections_by_principal()` snapshot and populates the new `DaemonStatus` field from it.
* **CLI** (`crates/astrid-cli/src/commands/who.rs`): renders one row per `(principal, connection)` from the new breakdown; falls back to the bare-count + default attribution when the new field is missing (older daemon).

5 files, +135/−54.

**Companion repo:** the cli capsule needs to take over publishing the lifecycle events stamped with the bound principal. Companion patch is on `unicity-astrid/capsule-cli` (`migrate/cargo-manifest` commit `c52c091`); separate PR.

## Test Plan

- [x] `cargo test -p astrid-kernel --lib` (88 passed)
- [x] Manual:
  - alice chat in flight → `who` renders `alice / cli / active`
  - default chat in flight → `who` renders `default / cli / active`
  - both disconnected → `who` renders `No clients connected.`

## Checklist

- [x] Linked to an issue (#738)


joshuajbouw:
Addressed in df5c2cca19e556c7e58c9a8f4e91c7bf76f06eb5 (rebased on origin/main; CHANGELOG cleaned).

**Suggestion (`who.rs` manual JSON traversal):** valid. Replaced the inline `{type, value}` unwrap with the shared `SocketClient::extract_kernel_response` helper that `ps.rs` and `daemon.rs` already use. Future changes to the IPC response envelope now propagate uniformly — no risk of `who` drifting from the canonical decoder.

PR #741, opened 2026-05-14
feat(cli/session): --session accepts UUID or name (auto-detect)

### LINKED ISSUE #740: feat(cli/session): --session should accept either a UUID or a name (auto-detect)
## Summary

`--print-session` reports a UUID, but feeding that UUID back into `--session` on the next invocation hashes it again into a fresh UUID v5 (NAMESPACE_URL), producing a different session. Operators who try the natural round-trip — copy the printed UUID and paste it into the next call — silently get a new session instead of resuming the prior one. The pattern that actually works (`--session NAME` for any non-UUID string) is undocumented in the help text.

## Motivation

Session round-trip is the primary multi-turn flow for headless usage (`astrid -p ... --session ...`). When the operator-facing API rounds-trips silently to "fresh session" instead of "resume", every batch script that interleaves `--print-session` and `--session` ends up running each turn in a new session — operators only notice when the LLM stops referencing context they thought it had.

`cargo`, `gh`, and `claude` all accept either a UUID or a name on the same flag for the same conceptual reason; this PR brings Astrid in line.

## Proposed implementation

`crates/astrid-cli/src/commands/headless.rs`: new `resolve_session_arg(s: &str) -> uuid::Uuid` helper. If `s` parses as a UUID, return it as-is (round-trip). Otherwise hash it via UUID v5 (NAMESPACE_URL) so the same name always maps to the same session.

`--print-session` output adjusts: a name input prints `name (uuid)` so the resolved form is visible; a UUID input prints just the UUID (already the resolved form). Both forms work for resume.

CLI help text rewrites under value name `ID_OR_NAME` to spell out the dual semantics.

## Verification

Manual:
- Turn 1 with `--print-session` captures uuid X.
- Turn 2 with `--session X` resumes turn 1's history (LLM recalls).
- Turn 3 with `--session some-name` starts a fresh named session.

joshuajbouw:
## Linked Issue

Closes #740.

## Summary

`--session` now auto-detects: an argument that parses as a UUID is used as-is for resume; anything else is treated as a stable session name and hashed via UUID v5. Matches `cargo` / `gh` / `claude` convention. Operators can copy the UUID printed by `--print-session` straight into `--session <that UUID>` for the next turn and get the same session — round-trip works.

## Changes

* `crates/astrid-cli/src/commands/headless.rs`: new `resolve_session_arg(s: &str) -> Uuid`. UUID input returned as-is; non-UUID input hashed via NAMESPACE_URL.
* `--print-session` shows `name (uuid)` for name input, just `uuid` for UUID input.
* `crates/astrid-cli/src/cli.rs`: help text rewritten under value name `ID_OR_NAME`.

2 files, +40/−14.

## Test Plan

- [x] `cargo build -p astrid`
- [x] `cargo test -p astrid --bin astrid -- --quiet` (194 passed)
- [x] Manual: `-p ... --print-session` captures UUID X; `-p ... --session X` resumes; `-p ... --session some-name` starts a fresh named session.

## Checklist

- [x] Linked to an issue (#740)


joshuajbouw:
Addressed in 1ce0b22922ad410b560ba85c2d0b099c1e5df6b6 (rebased on origin/main; CHANGELOG cleaned).

**Suggestions (`resolve_session_arg` returns redundantly-parseable bit + redundant `parse_str` in caller):** valid and applied. `resolve_session_arg` now returns `(Uuid, bool)` where the bool is `true` iff the input parsed as a UUID directly. `run_headless` destructures `(id, was_uuid)` and uses `was_uuid` for the `--print-session` branch instead of re-running `Uuid::parse_str(s)` on the same input. `run_snapshot_tui` takes `.0` since it doesn't care about provenance.

**Suggestion (move to shared utility module):** declining. There are currently two callers — both in this file — and only one differentiates UUID vs name. Lifting to a util module ahead of a third caller would be premature abstraction; if/when a chat-side resume path appears that needs the same convention, that's the moment to promote it.

PR #742, opened 2026-05-18
fix(sandbox)!: refuse silent bypass + ship AppArmor profile and astrid setup (closes #655)

### LINKED ISSUE #655: Sandbox-unavailable fallback in #651/#648 silently weakens subprocess-tier isolation — contradicts README's "Two sandboxes" promise
### Summary

## Summary                                                                                                                                                                            
                                                                                                                                                                                        
  PR #651 (merged as the fix for #648) introduced an AppArmor/bwrap-probe fallback: if `bwrap --unshare-user` fails — which is the default state on Ubuntu 24.04+ because               
  `kernel.apparmor_restrict_unprivileged_userns=1` is shipped enabled — `sandbox_prefix()` returns `None` and the caller in `astrid-mcp/src/server.rs` runs the native subprocess       
  **unsandboxed**, with a single `tracing::warn!` line as the only signal.                                                                                                              
                                                                                                                                                                                        
  This is a deliberate availability-over-security tradeoff, and I understand the reasoning — the alternative is "fresh Ubuntu 24.04 install silently can't launch MCP servers," which is
   worse UX. But the tradeoff isn't surfaced to the end user, and it contradicts what the README currently promises about capsule isolation.                                            
                                                            
  ## Why this matters: README vs. observed behavior                                                                                                                                     
                                          
  The README's **"Two sandboxes"** section reads:                                                                                                                                       
                                                                                                                                                                                        
  > **WASM sandbox.** Capsules run in WebAssembly via Extism/Wasmtime. No syscalls, no file descriptors, no host memory access. Every external resource (filesystem, network, IPC, KV
  storage) is gated behind a capability-checked host function. …                                                                                                                        
  >                                                         
  > **VFS overlay.** The agent operates against a copy-on-write filesystem. …                                                                                                           
                                              
  The "capsule" concept in the README is pitched as uniformly isolated. But capsules are actually serviced by **two different sandbox mechanisms** depending on the engine:             
                                                            
  | Capsule engine | Isolation mechanism | Affected by #651 fallback? |                                                                                                                 
  |---|---|---|                                                                                                                                                                         
  | Pure WASM (most `@unicity-astrid/capsule-*`) | Wasmtime memory sandbox + host ABI + 5-layer security interceptor | No — unaffected, still strong |                                  
  | MCP / native subprocess (`rmcp`) | Linux bwrap / macOS Seatbelt | **Yes — silently runs unsandboxed on 24.04+ default** |                                                           
  | OpenClaw Tier 2 (Node.js) | Same as MCP — native subprocess | **Yes — same** |
                                                                                                                                                                                        
  When the process sandbox is skipped, a Node.js MCP server can just call:                                                                                                              
                                                                                                                                                                                        
  ```js                                                                                                                                                                                 
  fs.readFileSync('/home/user/.ssh/id_rsa')                                                                                                                                             
  fs.writeFileSync('/home/user/.bashrc', malicious_content)                                                                                                                             
  ```                                                                                                                                                                                   
                                                            
  …directly against the real host filesystem. These calls never touch the Astrid host ABI, so **none** of the README's 5-layer security flow (policy → token → budget → approval →      
  audit) fires. The `--tmpfs ~/.astrid` overlay that #648 Bug 2 was specifically about hiding secrets from? Gone. The capability system that "lets trust expand gradually as the agent  
  proves reliable"? Irrelevant — the capsule isn't asking for capabilities, it's reading the file directly.
                                                                                                                                                                                        
  The upshot: a user on a default Ubuntu 24.04+ install who runs `cargo install astrid && astrid init` and installs any MCP / OpenClaw Tier 2 capsule gets a silent, un-isolated runtime
   for those capsules — while reasonably believing, based on the README, that capsules are sandboxed.                                                                                   
                                                            
  ## Repro of the silent-degrade signal                                                                                                                                                 
                                          
  ```bash                                                                                                                                                                               
  # On a fresh Ubuntu 24.04 install with default AppArmor settings:                                                                                                                     
  cat /proc/sys/kernel/apparmor_restrict_unprivileged_userns
  # -> 1                                                                                                                                                                                
                                                            
  # Install astrid (per README)                                                                                                                                                         
  cargo install astrid                                                                                                                                                                  
  astrid init                                                                                                                                                                           
                                                                                                                                                                                        
  # Any MCP / subprocess-tier capsule operation now spawns its backing process                                                                                                          
  # WITHOUT bwrap. The signal is a single tracing::warn! line in
  # ~/.astrid/log/astrid.YYYY-MM-DD.log of the form:                                                                                                                                    
  #   WARN ... bwrap sandbox unavailable: user namespace creation failed. ...
  #   Capsules will run without OS-level sandboxing.                                                                                                                                    
  ```                                                                                                                                                                                   
                                                                                                                                                                                        
  No prompt, no confirmation, no TUI indicator. Users who don't read logs won't know.                                                                                                   
                                                                                                                                                                                        
  (I cannot fully repro the end-to-end chain right now because #654 blocks the agent loop on released v0.5.1 + astralis@0.1.0 — the 5 core capsules fail manifest validation and the TUI
   hangs silently. But the sandbox-fallback path itself is straightforward to inspect in `crates/astrid-workspace/src/sandbox/bwrap.rs` — `bwrap_available()` + the caller in           
  `crates/astrid-mcp/src/server.rs`.)                       
                                                                                                                                                                                        
  ## Suggested directions                                   
                                                                                                                                                                                        
  One or a combination of:                                  
                                          
  1. **Policy becomes a first-class config choice.** Instead of `sandbox_prefix() -> Option<SandboxPrefix>` implicitly expressing "sandbox or fallback," introduce a tri-state config:
  `sandbox.policy = "required" | "preferred" | "off"`. Default to `"required"` so the astrid daemon *refuses* to launch subprocess capsules without a working sandbox — and surfaces a  
  clear error pointing the user at the `apparmor_restrict_unprivileged_userns` sysctl with the `sudo sysctl -w` command. Users who explicitly accept the risk (CI, development,
  constrained environments) can set `"preferred"` or `"off"` in `~/.astrid/config.toml`.                                                                                                
                                                            
  2. **Loud user-visible warning on every unsandboxed launch.** If fallback is allowed, surface it in the TUI (not just the log) — a banner/badge on the chat window like `⚠ Subprocess 
  capsules are running un-isolated. See: astrid sandbox status`. A log line visible only to operators who think to `tail` a file is not enough for a property the README promises.

  3. **README disclosure.** Either way, the "Two sandboxes" section should mention that the process-sandbox tier depends on `bwrap` + unprivileged user namespaces, that Ubuntu 24.04+  
  blocks this by default, and what the behavior is when unavailable. A one-paragraph note with the sysctl workaround would close the promise-vs-behavior gap.
                                                                                                                                                                                        
  4. **Orthogonal process-side idea.** Consider packaging a user-namespace-allowing AppArmor profile (e.g. `/etc/apparmor.d/astrid-userns` referencing `abstractions/userns` on systems 
  that support it). Debian packaging convention for tools that legitimately need user namespaces. Longer-term, but it'd let astrid work out of the box on hardened Ubuntu without the
  sysctl toggle *or* the fallback.                                                                                                                                                      
                                                                                                                                                                                        
  ## Related
                                                                                                                                                                                        
  - #648 — the OpenClaw Tier 2 / bwrap mount-ordering bug whose fix introduced this fallback
  - #651 — the fix PR (merged)            
  - #653 — tracking: production multi-tenancy. Different layer (per-principal isolation inside the daemon vs. OS-level process isolation), but thematically both surface that Astrid's
  isolation story has gaps that aren't user-visible.                                                                                                                                    
  - #654 — blocks end-to-end verification of #651 on the released path
                                                                                                                                                                                        
  ## Severity                                                                                                                                                                           
                                                                                                                                                                                        
  Not a runtime blocker — nothing breaks. This is a **security-story consistency** issue: what the README says and what Astrid does diverge on a common Linux distribution. Matters when
   someone audits the security model, less so day-to-day.   
                                                                                                                                                                                        
  ## Files / refs                                                                                                                                                                       
                                          
  - Fallback implementation: `crates/astrid-workspace/src/sandbox/bwrap.rs` (`bwrap_available()`)                                                                                       
  - Caller that consumes the `Option`: `crates/astrid-mcp/src/server.rs`                                                                                                                
  - README section: "Two sandboxes" (discuss `WASM sandbox` + `VFS overlay`, implicitly treats capsules as uniformly isolated)
  - AppArmor default: Ubuntu 24.04+ ships `kernel.apparmor_restrict_unprivileged_userns=1`                               


### Expected Behavior

Stated in README functionality should match actual behavior.

### Environment

Ubuntu 24.04

joshuajbouw:
## Linked Issue

Closes #655

## Summary

Three-commit branch that closes #655 — the silent sandbox bypass on Ubuntu 24.04+ that contradicts the README's "subprocess capsules are sandboxed" promise.

## Changes

- **6eda044 fix(workspace,mcp)!** — `SandboxPolicy` bi-state (`Required` default, `Off` escape hatch). `ProcessSandboxConfig::sandbox_prefix()` returns `Err` instead of `Ok(None)` when the sandbox can't be applied, and the MCP server-startup path propagates the error so the daemon refuses to launch the subprocess. `ASTRID_SANDBOX_POLICY=off` is the only escape hatch (trusted dev/CI). No "warn and fall through" middle state — that soft fallback is the bug.
- **a54b5a9 test(workspace)** — `sandbox_probe` example + Ubuntu 24.04 Docker rig used to verify the matrix of behaviours.
- **9aac6db feat(cli)** — `dist/apparmor/astrid` (narrow `flags=(unconfined)` profile granting only the `userns,` rule) + `astrid setup` subcommand that diagnoses the prerequisites and prints — or with `--apply` runs via `sudo` — the exact commands to install the profile at `/etc/apparmor.d/astrid`. With the profile loaded, `bwrap` regains `unshare(CLONE_NEWUSER)` on Ubuntu 23.10+/24.04 hosts where `kernel.apparmor_restrict_unprivileged_userns=1` is enabled — without requiring a host-wide sysctl flip (which would weaken every other unprivileged process on the host, not just Astrid).

### Why an AppArmor profile beats flipping the sysctl

The sysctl `kernel.apparmor_restrict_unprivileged_userns=0` re-enables unprivileged `CLONE_NEWUSER` for **every** process on the host. A path-scoped AppArmor profile (`profile astrid /usr/{,local/}bin/astrid flags=(unconfined)`) carves out the exception narrowly: host-wide policy stays intact, only the Astrid binary (and the children it execs, like `bwrap`) can create user namespaces. `flags=(unconfined)` means the profile only *adds* the `userns,` rule; the actual sandboxing happens inside the user namespace via bwrap.

### Automated

- [x] `cargo build -p astrid` clean on macOS
- [x] `cargo build -p astrid` clean on Ubuntu 24.04 (Docker)
- [x] `cargo test -p astrid --bins setup::` — 2 pass
- [x] `cargo clippy -p astrid --bins` clean
- [x] `apparmor_parser -p dist/apparmor/astrid` parses on Ubuntu 24.04
- [x] `astrid setup --print-apparmor | apparmor_parser -p` parses on Ubuntu 24.04
- [x] Unit tests on `setup.rs` pin the profile invariants (`userns,`, `flags=(unconfined)`, profile name `astrid`) so future edits can't regress the fix

### Manual

1. On an Ubuntu 24.04 host with `kernel.apparmor_restrict_unprivileged_userns=1`:
   ```
   astrid setup                # diagnoses, prints commands
   sudo astrid setup --apply   # installs profile via sudo
   astrid setup                # bwrap probe now passes
   ```
2. Trigger a subprocess capsule launch and confirm it runs sandboxed (was failing-closed before `setup`, now succeeds).

Live install on an Ubuntu 24.04 host is deferred to operator validation — Docker can't host-enforce AppArmor from inside a container, so the end-to-end `apply → load → probe-passes` cycle needs a real host.

## Checklist

- [x] Linked to an issue (#655)
- [x] CHANGELOG.md updated under `[Unreleased]` (`### Added`, `### Breaking`)


joshuajbouw:
Addressed all review feedback + the Ubuntu CI failure in 5d3ea84.

## CI failure (root cause)
6 `astrid-mcp` tests were calling `build_sandboxed_command` and depending on the silent fallback that #655 explicitly removed. Under `SandboxPolicy::Required` they now panic at `.expect()` on hosts without working unprivileged userns (the GitHub Ubuntu runner). Fix: threaded a `sandbox_policy_override` through `ServerManager` (`.with_sandbox_policy()`) so tests can opt into `Off`.

- 5 tests (`test_sandboxed_command_clears_env`, `_blocks_dangerous_env`, `_writable_root_priority_cwd_first`, `_workspace_second`, `test_build_sandboxed_command_resolves_absolute_binary_path`) → opt into `Off`; they exercise env scrubbing / writable-root precedence / binary resolution, not the sandbox-policy decision.
- 1 test (`test_build_sandboxed_command_adds_sandbox_prefix`) → keeps the default `Required` policy and gates on a runtime bwrap probe; it skips with a clear message when the sandbox can't be applied (instead of dishonestly testing the old fallback path that #655 removed).

The dropped `if-sandbox-unavailable` conditional branches in those tests were checking for the exact silent-fallback behaviour this PR removes — keeping them would have invited the regression back in.

## Review comments

### High — hardcoded `/tmp/astrid-apparmor-profile`
You're right, that's a classic symlink-attack surface. Replaced:
- **Auto-apply path** (`astrid setup --apply`): now stages through `tempfile::NamedTempFile` (random name, O_EXCL).
- **Operator copy-paste recipe**: now uses `TMPFILE=$(mktemp -t astrid-apparmor.XXXXXX)` and quotes the path everywhere.
- Added `install_commands_use_mktemp_not_predictable_tmp_path` unit test that pins `mktemp` and forbids the legacy predictable path so a future edit can't quietly regress this.

### Medium — stale `Preferred` doc references (5 sites)
All five fixed:
- `crates/astrid-workspace/src/sandbox/mod.rs:349,418,470`
- `crates/astrid-mcp/src/server.rs:544,565`

(That variant was dropped on review during the original push — `SandboxPolicy` is binary now: `Required` default, `Off` escape hatch. Apologies for the staleness.)

### Medium — unused `_cmds` parameter on `apply_install`
Removed; the function builds its own `tempfile` source so the parameter was decorative.

PR #804, opened 2026-05-29
ci: cancel superseded CI runs for PRs

### LINKED ISSUE #803: ci: cancel superseded CI runs for PRs (rust-cache split out)
## Summary

CI reruns are more expensive than they need to be. Two fixes to `ci.yml`:

1. **No concurrency cancellation.** `ci.yml` has no `concurrency` block, so every push to a PR runs to completion even when a newer push (a review fix, a rebase) obsoletes it. Add `cancel-in-progress` for PR runs (let pushes to `main`/`stream-*` each finish so every commit keeps a record).

2. **Effectively-frozen Rust cache.** The manual cache key is `${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}`, but core's `Cargo.lock` is gitignored — so in CI `hashFiles` matches nothing and the key collapses to the constant `${{ runner.os }}-cargo-`. `actions/cache` doesn't re-save on an exact-key hit, so after the first run the `target/` cache freezes and later runs restore stale/partial artifacts. Switch the cargo/target caches to `Swatinem/rust-cache@v2` (already used in `release.yml`), which keys on lockfile + rustc + job and prunes `target/`. The QuickJS-kernel `actions/cache` step stays as-is (rust-cache doesn't cover arbitrary artifacts).

Net: faster, cheaper reruns — including the unavoidable post-rebase reruns — with no change to coverage.

## Notes

- Docs/changelog-only changes already skip the heavy jobs (`ci.yml` is path-filtered to `**.rs` / `Cargo.*` / `.github/workflows/**`), and `main` is unprotected, so there's no required-check deadlock to handle.
- CI-only change; no runtime behaviour changes, so no CHANGELOG entry (and `changelog.yml`'s path filter does not require one).

joshuajbouw:
## Linked Issue

Closes #803

## Summary

Cancel superseded CI runs on PRs, so a review fix or a rebase doesn't leave an obsolete run burning minutes to completion. This is the part of the original idea that's an unambiguous win.

## Changes

- Add a `concurrency` group (`${{ github.workflow }}-${{ github.ref }}`) with `cancel-in-progress` for `pull_request` events only. Pushes to `main`/`stream-*` are **not** cancelled, so every landed commit keeps its own run.

## Dropped: rust-cache (and why)

The first cut also swapped the manual `actions/cache` for `Swatinem/rust-cache@v2`. That regressed the heavy **Test** job: the old constant-key cache was restoring a warm `target/` that kept the link of the big test binaries (wasmtime/oxc/cranelift) + the QuickJS `js-pdk` auto-build under the ubuntu runner's ~14 GB disk limit. rust-cache started cold (and `CACHE_ON_FAILURE: false` keeps re-runs cold), forcing a from-scratch build that crashed the linker — `ld terminated with signal 7 [Bus error], core dumped`. Reverted to the existing cache.

The cache/disk fragility is real (the Test job runs near the disk edge on `main` too — the "stale" frozen cache is load-bearing), but it's pre-existing and wants a disk-hardened fix (e.g. a free-disk-space step) on its own, not rushed onto a "faster CI" PR.

## Test Plan

CI-configuration-only change; the meaningful validation is this PR running CI on its own updated config (it edits `.github/workflows/**`). YAML validated locally with `yaml.safe_load`.

### Automated

- [x] `cargo test --workspace` passes (unchanged — no Rust touched)
- [x] No new clippy warnings (unchanged — no Rust touched)

## Checklist

- [x] Linked to an issue
- [ ] CHANGELOG.md updated under `[Unreleased]` — N/A: CI-only, no runtime change

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,13 @@ on:
       - '**/Cargo.lock'
       - '.github/workflows/**'
 
+concurrency:
+  # Cancel superseded in-progress runs for the same PR (a review fix or a
+  # rebase makes the prior run obsolete). Pushes to main / stream-* are not
+  # cancelled, so every landed commit keeps its own run on record.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: 1
@@ -41,16 +48,8 @@ jobs:
         with:
           node-version: "22"
 
-      - name: Cache cargo registry
-        uses: actions/cache@v4
-        with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-            target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-cargo-
+      - name: Cache Rust build
+        uses: Swatinem/rust-cache@v2
 
       - name: Run cargo check
         run: cargo check --workspace --all-features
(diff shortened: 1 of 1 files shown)


joshuajbouw:
Closing as its problematic, and doesn't yield much benefits right now.

Issue #813, opened 2026-06-01
[Epic] Capsule orchestration path collapses under concurrent prompts (1000-principal smoke)


joshuajbouw:
## Verdict

A 1000-principal scale smoke test exposes a sharp concurrency cliff on the gateway → bus → react → prompt-builder → router → openai-compat path. The static side of the 1000-principal target passes — sub-linear memory (3.7 KB RSS/principal), single WASM instance shared across 1000+ principals, multi-turn isolation correct, zero `CapabilityDenied` — but the runtime under any meaningful concurrency does not.

## Reproduction

Fresh ASTRID_HOME, daemon booted with the v0.7 chain of recent fixes applied (the kernel principal-propagation fix, the per-invocation env-overlay fix, plus the corresponding capsule-cli / capsule-registry / capsule-react fixes). Fake OpenAI-compatible LLM server bound at `127.0.0.1:1235` so the LLM side is effectively instant.

1. Mint 1000 principals through `/api/auth/redeem` at concurrency 20 — completes in 55s, zero failures.
2. Fire 10 sequential prompts across 10 randomly-chosen principals — all succeed; p50 ≈ 4.9s end-to-end (the fake LLM responds in <1ms; the remaining ~5s is pure Astrid orchestration overhead).
3. Fire 100 concurrent prompts across 100 randomly-chosen principals (xargs -P 100, distinct session_ids).

## Observed at step 3

- **100 / 100 requests timed out** at `curl --max-time 30`. The gateway accepted each request and emitted the `event: ready` SSE frame (~114 bytes), then the stream stalled.
- **Only 1 of 100** ever reached the fake LLM during the load window. The other 99 never made it through the capsule chain.
- A single prompt fired immediately *after* the load completed returned in ~4s — confirming the daemon was alive and the pipeline wasn't dead, just gridlocked under contention.

p50/p95/p99/max of ~30s are all `--max-time` artifacts, not real latencies. True per-request latency under 100-wide load is unbounded.

## Correlated kernel signals during the load

Counted from the daemon log over the 87s test:

- **15× `ipc::recv: mixed-principal batch truncated to first publisher's messages` (security_event=true)** with `kept=1 dropped=6` typical shape. The kernel's per-recv homogeneous-principal gate is doing what it should — but the truncation drops ~6 messages each time, almost certainly causal for the client timeouts.
[...]

Issue #69, opened 2026-02-18
docs: document two-channel design for McpPlugin vs McpClient notifications


joshuajbouw:
McpPlugin.load() creates its own mpsc channel for ServerNotice, separate from McpClient. The McpClient listener silently ignores ConnectorsRegistered. This is correct but should be documented for future maintainers.

joshuajbouw:
The current architecture has no McpPlugin abstraction or the two-channel contract described here. McpClient owns the remaining notice path, and documenting the removed split would make the current design less clear. Closing this as superseded.

PR #841, opened 2026-06-05
fix(dispatcher): never hand back a closed per-topic consumer (burst stall)

### LINKED ISSUE #837: Dispatcher drops events under burst when a per-topic consumer is reclaimed
Under a concurrent burst, a per-(capsule,topic) dispatch consumer can leave a closed sender in the queue map (idle-evict race or an ended task); get_or_spawn_consumer returned it, so every later send failed Closed and the event was dropped, permanently stalling that route (e.g. react/user.v1.prompt) while the capsule stayed healthy. Skip closed senders and re-spawn. Surfaced by a live 100-wide prompt smoke (response_seen 4/100).

joshuajbouw:
## Linked Issue
Closes #837

## Summary
Under a concurrent prompt burst the per-(capsule,topic) dispatch consumer for a route could leave a CLOSED sender in the queue map (idle-evict TOCTOU, or a consumer task that ended). `get_or_spawn_consumer` returned that dead sender, so every later `try_send` failed `Closed` and the event was dropped permanently, stalling that route (e.g. `react`/`user.v1.prompt`) while the capsule stayed healthy. Pre-existing on `main` (#828 foundation); surfaced by the Phase 2 live smoke. Independent of the Phase 2 stack.

## Changes
- `get_or_spawn_consumer` now skips a mapped sender whose receiver is gone (`is_closed()`) and re-spawns (the `insert` overwrites the stale entry) — never returns a closed sender.
- `dispatch_single` keeps a re-spawn-and-retry backstop for the narrow return-to-send window, and distinguishes `Full` (intended bounded shed-load, recoverable via the requester timeout) from `Closed` (a bug, never dropped).
- Deterministic regression test: a dispatch through a pre-seeded closed mapped sender must re-spawn and deliver.
- CHANGELOG.

### Automated
- [x] 19 dispatcher tests incl. the new closed-sender regression; clippy `-D warnings`, fmt
### Manual
- [x] Live-verified on the same 100-wide smoke that exposed it: response_seen 4/100 -> 95/100; zero channel-closed drops (39 before, 0 after); recovers post-burst (26/30 vs 2/30).

## Checklist
- [x] Linked to an issue

Refs #69, #813.


joshuajbouw:
Addressed in e561354:
- **stale closed senders (gemini, high):** `get_or_spawn_consumer` now **removes** a closed mapped entry rather than relying on the later insert to overwrite it. The skip-only approach leaked in the degrade-to-shared path — the insert re-keys to `(capsule, None)` and never overwrites the original `(capsule, Some(principal))` dead entry, so the `Sender` + `PrincipalKey` string accumulated and slowed `queues_per_capsule`. Both the per-principal and shared-key checks now remove-on-closed.
- **retry log (Copilot):** `dispatch_single`'s re-spawn retry distinguishes `Full` (intended shed-load drop) from `Closed` (a bug immediately after a fresh re-spawn) — the latter is now logged as a `security_event`, not folded into the benign backpressure log.
- **CHANGELOG (Copilot):** merged the dispatcher bullet into the existing `### Fixed` under `[Unreleased]` instead of a duplicate heading.

dispatcher tests green (19); CI-exact clippy + fmt clean.

PR #844, opened 2026-06-05
fix(gateway): per-IP rate-limit pair-device/redeem

### LINKED ISSUE #843: gateway: POST /api/auth/pair-device/redeem has no per-IP rate limiter
## Summary

`POST /api/auth/pair-device/redeem` is an unauthenticated, public route — the pair-token is the auth, and the kernel's admin dispatcher bypasses the capability gate for `PairDeviceRedeem`. Unlike its sibling `POST /api/auth/redeem` (which is per-IP rate-limited via `redeem_limiter` in `routes/auth.rs::post_redeem`), pair-redeem has **no** brute-force fence in front of the kernel's constant-time pair-token scan (`pair_device_handlers.rs`). An attacker can enumerate pair-tokens as fast as the network allows.

Surfaced by the CLI/API sysadmin parity audit (the one externally-facing security asymmetry between the two unauthenticated redeem routes).

## Impact

- Public, unauthenticated route with no per-IP throttle.
- Only backstop today is the kernel token scan + single-use deletion.
- Security: medium (token space is large, but the missing throttle is an unnecessary asymmetry vs invite-redeem).

## Fix

Apply the same per-IP throttle `post_redeem` uses, **sharing** the existing `redeem_limiter` so the per-IP budget is spent across both unauthenticated redeem routes (can't be dodged by alternating routes). Honour `X-Forwarded-For` only behind a configured trusted proxy (`resolve_client_ip`). Document the 429 in the route's OpenAPI contract.

## Follow-up (not in this issue)

`astrid_gateway_redeem_attempts_total` / `astrid_gateway_redeem_rate_limited_total` are declared in `metrics.rs` but not currently incremented by either redeem handler — wiring them would make the throttle observable. Tracked separately.

joshuajbouw:
## Linked Issue

Closes #843

## Summary

`POST /api/auth/pair-device/redeem` is an unauthenticated, public route (the pair-token is the auth; the kernel cap-gate is bypassed for `PairDeviceRedeem`). Unlike its sibling `POST /api/auth/redeem`, it had **no per-IP rate limiter** in front of the kernel's constant-time pair-token scan, so pair-tokens were brute-forceable at network speed. This was the one externally-facing security asymmetry between the two unauthenticated redeem routes flagged by the CLI/API sysadmin parity audit.

## Changes

- `routes/auth.rs::post_pair_device_redeem` now runs the same per-IP throttle as `post_redeem`, **sharing the one `redeem_limiter`** so the per-IP budget is spent across both unauthenticated redeem routes and can't be dodged by alternating between them.
- Adds the `ConnectInfo<SocketAddr>` + `HeaderMap` extractors and `resolve_client_ip` (honours `X-Forwarded-For` only behind a configured trusted proxy).
- Documents the `429` (`retry_after_secs`) in the route's OpenAPI `responses`.
- New `tests/router.rs` test: pre-seeds the shared limiter and asserts a pair-redeem from that IP is rejected with `429` before reaching the daemon.
- CHANGELOG `### Security` entry.

## Test Plan

- `cargo test -p astrid-gateway` — 13/13 pass incl. the new `pair_device_redeem_is_rate_limited_per_ip`.
- `cargo clippy -p astrid-gateway --all-targets -- -D warnings` — clean.
- `cargo fmt -p astrid-gateway --check` — clean.

No contract change beyond the added `429` response (the route already returned `429`-class behaviour nowhere; this adds it). Behaviour for legitimate single-attempt pairing is unaffected (default interval 1s).

> Note: `astrid_gateway_redeem_attempts_total`/`_rate_limited_total` are declared but not incremented by either redeem handler today; wiring them is noted as a follow-up in #843.


joshuajbouw:
Addressed in 40f10c2:
- **retry-after rounding (gemini, medium):** `retry_after_secs` now rounds **up** (saturating-add the sub-second remainder) so a strict `Retry-After` client never retries before the window elapses (4.5s → 5, not 4). Applied to both redeem routes for consistency.
- **OpenAPI status codes (Copilot):** both redeem routes now document **502** for a kernel-rejected token (`GatewayError::Kernel` → `BAD_GATEWAY`) and scope **500** to upstream-unreachable / unexpected-shape, instead of conflating them under 500. (`post_redeem` had the same pre-existing inaccuracy, so both are corrected.)

clippy `-D warnings` + fmt clean; the per-IP rate-limit router test passes.

PR #846, opened 2026-06-05
fix(kernel): audit failed token redeems with a failure outcome

### LINKED ISSUE #845: kernel: failed token redeems are audit-stamped as success (audit-outcome fidelity)
## Summary

`InviteRedeem` and `PairDeviceRedeem` bypass the capability preamble in `kernel_router/admin/mod.rs::handle_admin_request` (the token is the auth — the redeemer's principal does not exist yet). The dispatcher special-cases them, but it records the audit row `AuthorizationProof::System` + `AuditOutcome::success` **before** dispatching the handler.

A redeem the handler then rejects (invalid / expired / consumed / forged token, or an internal store error) still leaves a **success-stamped** `AdminRequest` audit row. The only failure signal is the handler's tracing `warn!(security_event=true)` — there is no failure-outcome audit row.

Surfaced by the CLI/API sysadmin parity audit as the one audit-fidelity gap.

## Impact

- A security team relying on **audit rows** (not tracing logs) to detect token brute-forcing cannot — failed/forged-token attempts look identical to successful redeems at the `AuditOutcome` level.
- The surrounding comment even claims "every redeem still produces an audit row (allow OR deny)", which the implementation never delivered.

## Fix

Dispatch first, then record the audit row with the **real** outcome: a rejected token (`AdminResponseBody::Error`) writes `Denied` + `Failure(reason)`; a mint writes `System` + `Success`. Extract the decision into a small pure helper (`redeem_audit_proof`) with a unit test. No WIT/contract change.

joshuajbouw:
## Linked Issue

Closes #845

## Summary

`InviteRedeem` / `PairDeviceRedeem` bypass the capability preamble (the token is the auth — the redeemer's principal does not exist yet). The admin dispatcher special-cased them and stamped the audit row `AuthorizationProof::System` + `AuditOutcome::success` **before** dispatching the handler, so a rejected token (invalid / expired / consumed / forged, or an internal store error) still left a **success** row. Failed/brute-force redeem attempts were therefore invisible in the audit log — only in a tracing `warn!(security_event=true)`. The one audit-fidelity gap found by the CLI/API sysadmin parity audit.

## Changes

- `kernel_router/admin/mod.rs`: the redeem branch now dispatches **first**, then records the audit row with the real outcome — a rejected token writes `Denied` + `Failure(reason)`; a mint writes `System` + `Success`. This is the "allow OR deny" the surrounding comment always promised.
- Extracted the decision into a small pure helper `redeem_audit_proof(&AdminResponseBody) -> (AuthorizationProof, AuditOutcome)`.
- New unit test `redeem_audit_proof_marks_rejected_token_as_failure` (admin/tests.rs) pins `Error → Denied + Failure` and `non-error → System + Success`.
- CHANGELOG `### Security` entry.

## Test Plan

- `cargo test -p astrid-kernel admin::` — 68/68 pass incl. the new test.
- `cargo test -p astrid-kernel --lib` — green (the only 2 failures are `socket::tests` `UnixListener::bind` `PermissionDenied` under the sandbox; both pass with the sandbox disabled).
- `cargo clippy --workspace --all-features -- -D warnings` (CI-exact) — clean for the lib helper.
- `cargo fmt -p astrid-kernel --check` — clean.

## Notes

- The audit row is recorded before `publish_response`, so a client that receives the redeem response can still query the audit log and find the row — no race introduced.
- Trade-off: if `handlers::dispatch` itself *panics* (vs returning `Error`), no row is written; a panic is an abnormal crash logged separately, and dispatch returns `Error` for all normal rejections.


joshuajbouw:
Addressed in c9f2868:
- **avoid clone (gemini, medium):** the redeem branch diverges with an early `return`, so `audit_params` can be **moved** there instead of cloned (the non-redeem allow/deny paths keep their clone/move pair). Saves one `Option<serde_json::Value>` allocation per redeem.

Not changed: Copilot claimed `matches!(req.kind, ...)` moves `req.kind` and would fail to compile. It doesn't — the `{ .. }` patterns bind nothing, so no move occurs; the kernel builds and `redeem_audit_proof` tests pass. Left as-is.

PR #890, opened 2026-06-10
feat(capsule): read-only file-injection for sandboxed spawns

### LINKED ISSUE #881: host: generic per-spawn read-only file injection (hash-pinned, agent-neutral sandbox primitive)
## Problem

When the host spawns a supervised `claude -p` agent (e.g. via the `sage` bundle's native-tools model), the per-principal tool/permission/sandbox posture is authored into `<home>/.claude/settings.local.json`. That is Claude Code's **Local** tier — the weakest, fully session-overridable tier in its precedence order (`Managed > CLI args > Local > Project > User`). A capable or prompt-injected session can edit `permissions.allow`, the `sandbox` block, or the deny list in that file and Claude will honour the edit.

The argv flags (`--sandbox`, `--permission-mode dontAsk`, `--disallowedTools …`) and the Astrid host OS-sandbox (bwrap/seatbelt) remain binding regardless, so this is **not** an escape of the host boundary. But the *fine-grained* posture (which exact tools/commands/domains are allowed, `allowUnsandboxedCommands:false`, `allowManagedPermissionRulesOnly`, `sandbox.network.allowManagedDomainsOnly`) is only as strong as the overridable Local tier.

The un-overridable tier is Claude Code's **Managed** settings, which load from a fixed SYSTEM path:
- macOS: `/Library/Application Support/ClaudeCode/managed-settings.json`
- Linux: `/etc/claude-code/managed-settings.json`

A WASM capsule with `fs_write` scoped to `home://` cannot author a system path, so the capsule cannot deliver un-overridable governance on its own.

## Proposed solution

In the host process-spawn path (`astrid-capsule` `engine/wasm/host/process/{managed,persistent}.rs` + `astrid-workspace` sandbox), **bind-mount a capsule-authored `managed-settings.json` (from the principal home) onto the canonical managed-settings path INSIDE the OS sandbox** that already wraps the agent process:
- Linux (bubblewrap): `--ro-bind <home>/.claude/managed-settings.json /etc/claude-code/managed-settings.json` — clean.
- macOS (Seatbelt): the canonical `/Library/Application Support/ClaudeCode/` path is the open question — Seatbelt has no bind-mount; options are a Seatbelt-visible redirect, a per-principal copy into a profile-allowed path, or a host-prepared overlay. **This is the design gap.**

With the bind-mount in place, the spawn-side posture (`sandbox.enabled`, `allowUnsandboxedCommands:false`, `permissions.deny`, `allowManagedPermissionRulesOnly`, `sandbox.network.allowManagedDomainsOnly`) becomes session-un-overridable.

## Scope / acceptance
- Linux bind-mount of a per-principal managed-settings.json into the canonical path inside the existing sandbox.
- macOS Seatbelt equivalent (or a documented decision on the redirect mechanism).
- The agent's spawn config selects which authored file to mount; absence is a no-op (no managed tier).

## Notes
The sandbox + argv floor already bound the agent regardless; this hardens the fine-grained policy from overridable to binding. The native-tools agent model is shipped capsule-side and is functional on the argv floor today — this issue is the hardening that closes the overridable-Local-tier gap.

joshuajbouw:
## Linked Issue

Closes #881

## Summary

Implements the generic, agent-neutral host primitive from #881: at process spawn, the host exposes host-verified, **read-only** bytes to a spawned child's existing OS sandbox (`bwrap` on Linux, Seatbelt on macOS). The motivating consumer is un-overridable per-spawn agent governance (a supervised agent reads a policy file a prompt-injected session cannot rewrite), but the host is **content-opaque** — it never parses the bytes, and it owns placement, so every agent-specific detail stays in the adapter and the kernel stays agent-neutral.

The capsule hands the host `content` bytes plus a **placement** describing how the child should find them. Two modes, both exposing the same verified bytes read-only:

- **`env-pointer(env-var)`** — the host materializes the snapshot at a host-owned path, exposes it read-only (Linux `--ro-bind P P`; macOS Seatbelt `file-read*` allow + a trailing `file-write*` deny on the literal), and sets the named env var on the child to that path. OS-agnostic (Linux + macOS); the host owns the path, so there's no caller-chosen target and no host write to a caller-named path. For agents whose un-overridable tier is reachable via an env-redirected file (Claude `CLAUDE_CODE_MANAGED_SETTINGS_PATH`, Gemini `GEMINI_CLI_SYSTEM_SETTINGS_PATH`).
- **`fixed-path(path)`** — the host `--ro-bind`s the snapshot at an absolute in-sandbox path. Linux-only (bwrap remap); rejected on macOS with `invalid-input` (no remap; a host write to a caller-named path would be an escalation). For agents whose enforced tier is a fixed path with no env redirect (Codex `/etc/codex/requirements.toml`).

## Changes

- **`astrid-workspace` sandbox**: `RoInjection` + `ProcessSandboxConfig::with_ro_inject`; Linux `--ro-bind` (after the writable bind, before `--unshare-all`); macOS Seatbelt `(allow file-read* (literal target))` + trailing `(deny file-write* (literal target))`. `SandboxCommand::wrap_with_injections` (no-injection path byte-identical to `wrap`); non-empty injections on an unsandboxed platform fail secure.
- **`astrid-capsule` `host/process/inject.rs`**: BLAKE3-hash `content`, snapshot to a private `TempDir` (host-owned, outside every VFS mount), **re-read and verify** against the pin (closes the copy→expose TOCTOU); per placement, emit the sandbox bind spec and (for `env-pointer`) the host-set env var. RAII guard (the `TempDir`) cleans up on the child's lifetime owner: a local for `spawn`, the `ManagedProcess` for `spawn-background`, the registry entry (reaped by value) for `spawn-persistent`. The spawn audit records `env:<var>=<hash>` / `path:<p>=<hash>`. The host sets the env-pointer var authoritatively on the child (not via the inert guest `spawn-request.env`).
- **WIT**: the `wit/` submodule is bumped to the file-injection commit and `wit-staging` is regenerated, so `bindgen!` produces `SpawnRequest.file_injections`, `FileInjection { content, placement }`, and `InjectionPlacement`.
- **Validation**: `env-pointer` var name non-empty / no `=` / no NUL; `fixed-path` absolute / no forbidden chars / no `..`; per-injection content cap (1 MiB → `too-large`); empty `file-injections` is a zero-overhead no-op.

### Automated

- [x] `cargo test --workspace` passes
- [x] No new clippy warnings (`cargo clippy --workspace --all-features -- -D warnings`)
- New unit tests: bwrap ro-bind positioning; macOS Seatbelt read-allow + trailing write-deny ordering; `wrap_with_injections` unsafe-path rejection; env-var-name and fixed-path (incl. `..`) validation; write+verify roundtrip + integrity-mismatch detection; 0600 snapshot mode; empty-injections no-op.

### Manual (optional)

Not applicable — no runtime behaviour change when `file-injections` is empty (every existing spawn); the injection path is exercised by the unit tests above.

## Dependencies / merge order

Third leg of a contract-first change — merge in order:

1. unicity-astrid/rfcs#22 (Host ABI RFC — "Read-only file injection")
2. unicity-astrid/wit#15 (canonical WIT)
3. **Re-pin** this PR's `wit/` submodule from the pre-merge branch commit `b8fdb6f` to the `wit@main` merge commit, then merge this.

The CI submodule ↔ `wit-staging` dirty-check enforces the pin stays consistent.

## Checklist

- [x] Linked to an issue (#881)


joshuajbouw:
Addressed the review (61ac328): `write_private` now uses `create_new` (O_CREAT | O_EXCL) instead of `create(truncate)`, so the snapshot write refuses any pre-existing entry or symlink at the destination rather than truncating through it — with a regression test. In practice the destination is always a fresh unique index under a host-private `TempDir` (unreachable by a guest), but this makes the freshness an enforced invariant.

The other two findings were already resolved in the amended commit: `reject_target_within_vfs` (macOS symlink bypass) is gone because non-Linux `FixedPath` is now rejected outright, and the unbounded-read/OOM concern doesn't apply — injected content is guest-supplied and capped at `MAX_INJECTION_BYTES` (1 MiB) before any fs work. Build + clippy + inject tests green.

Issue #852, opened 2026-06-05
feat(kernel): own principal attribution via per-connection session binding (supersede per-message publish-as)


joshuajbouw:
## Summary

`astrid:ipc/host.publish-as(topic, payload, principal)` lets a capsule publish a bus message stamped with a *chosen* principal instead of its own. It is the mechanism the CLI/uplink proxy uses to forward an authenticated socket client's principal onto the kernel's admin / capability-gating path.

This is **not a hole today** — it is gated and audited (see Current state). This issue proposes the longer-term evolution: move principal attribution into **core**, so identity stops being something a capsule re-asserts per message.

## Current state (acceptable, keep for now)

- **Gated:** `publish_as` returns `CapabilityDenied` unless the caller holds the uplink/`net_bind` capability. Only the one uplink (capsule-cli, the socket authority) can call it; a regular capsule cannot impersonate.
- **Audited:** every call emits an `astrid:ipc/host.publish-as` audit row.
- **Trusted, not verified:** the kernel trusts the uplink's stamp — the uplink owns the transport/handshake and is assumed to have authenticated that principal. So `publish-as` is a *per-message* principal assertion by a trusted uplink.

## Proposal

Identity is a kernel concern (the kernel is the trust authority). The catch is that the kernel cannot authenticate transport clients itself — the uplink owns the socket — so the uplink must assert transport-identity *somehow*. The cleaner shape is to change the **granularity** rather than delete the mechanism:

- **Per-connection session binding** instead of per-message `publish-as`: the uplink authenticates a client **once**, the kernel records `connection -> principal`, and subsequent messages on that connection auto-attribute. One assertion per session instead of per message shrinks the trust surface — a buggy or compromised uplink can no longer vary the principal mid-stream — and lets the kernel own attribution.
- **Provenance — forward-not-forge:** a capsule publishing as principal P is legitimate when it is *forwarding* a message it actually received from P, but not when it originates a P it never saw. A future design can make delegation traceable rather than blanket-trusted.
- Once attribution is connection-bound, `publish-as` can likely be **deprecated** (or narrowed to the forwarding case).

## Relationships

- Builds on the connection-lifecycle work (rfcs#29) — it already models connect/disconnect, the natural place to bind `connection -> principal`.
- Aligns with kernel-is-dumb: identity/gating is the kernel's job; the per-message re-assertion currently living in capsule-space is the thing to reclaim.

## Out of scope / non-goals

[...]

PR #896, opened 2026-06-10
feat(cli): add global --principal flag with per-connection stamping

### LINKED ISSUE #895: feat(cli): process-wide --principal flag with per-connection stamping
## Summary

The CLI currently attributes work via per-message principal stamps resolved ad hoc (active-agent context for admin commands, absent for most other sends). As a step toward per-connection session binding (#852), the CLI should act as exactly **one** principal for its whole process lifetime, declared up front and stamped consistently on every IPC message.

## Scope

- Global `--principal <NAME>` flag + `ASTRID_PRINCIPAL` env var; precedence: flag > env > active-agent context > `default`.
- Resolve + validate once at startup via `PrincipalId::new` (clear error naming the constraint).
- `SocketClient` binds the principal at connect time and stamps any outbound message that doesn't carry an explicit one — a single seam covering every CLI send path.
- Counterpart: the uplink proxy (capsule-cli) pins the first principal seen on a connection, auto-attributes unstamped messages to it, and drops conflicting stamps.

Gateway is out of scope (it stamps the verified bearer principal already).

joshuajbouw:
## Linked Issue

Closes #895 (step toward #852 per-connection session binding)

## Summary

The CLI now acts as exactly one principal for its whole process lifetime. The principal is resolved once at startup — precedence `--principal` flag > `ASTRID_PRINCIPAL` env > active-agent context (`astrid agent use`) > `default` — validated through `PrincipalId::new`, and stamped on every IPC message the process sends. The uplink proxy's counterpart change (capsule-cli) pins the first principal per connection and drops mismatches, so a consistent stream is now a correctness requirement, not a courtesy.

## Changes

- `crates/astrid-cli/src/cli.rs`: global `--principal` flag (`env = "ASTRID_PRINCIPAL"`).
- `crates/astrid-cli/src/principal.rs` (new): pure precedence/validation resolver + process-wide `OnceLock`; invalid input exits 1 naming the constraint.
- `crates/astrid-uplink/src/socket_client.rs`: `SocketClient::connect` now binds a principal; `send_message` stamps it on any message without an explicit one — one seam covering all CLI send paths. Explicit stamps (emit, mcp-watch publish-as attribution) win.
- `AdminClient`/`KernelClient` feed their `caller` into the binding; CLI call sites use the process principal.
- Gateway untouched (does not use `SocketClient`; stamps verified bearer principals as before).

Default behaviour is preserved: the kernel router already collapses absent → `PrincipalId::default()` (existing test at `kernel_router/mod.rs:921`), so messages that previously omitted the principal and now carry `default` behave identically.

## Test Plan

- Unit tests: precedence resolution (flag > env > default), invalid-principal rejection with constraint-naming message, boundary lengths.
- CI gates: check/fmt/clippy/test workspace, all green.
- Verified live against a real daemon (isolated `ASTRID_HOME`): `--principal alice` flows end-to-end (connection binds alice, kernel capability gate keys off alice, results demux back to the alice-bound client); invalid principal errors before any connection; unprovisioned principal fails closed at the kernel gate.


joshuajbouw:
Addressed the principal-binding review (4f7cc38): `watch.rs` (both sites) and `astrid-emit` now bind the uplink to the resolved principal via `PrincipalId::new` instead of `PrincipalId::default()`. As the `principal` module itself documents, the proxy pins the first principal per connection and drops mismatched stamps — so a default-bound connection would silently drop the explicitly-stamped enumerate/hook messages for any non-`default` principal. `watch::run` treats an invalid principal as non-fatal (logs, disables live-reload pushes); enumerate/emit propagate. Build + clippy + emit lib tests green.

PR #907, opened 2026-06-15
fix(capsule-install): materialize home://wit mirror so introspection tools work

### LINKED ISSUE #906: introspection: home://wit is never populated so list_interfaces/read_interface fail
## Summary

The system capsule (`capsule-system`) exposes two LLM introspection tools, `list_interfaces` and `read_interface`, that read WIT interface contracts from `home://wit/` (constant `WIT_DIR = "home://wit"`). That directory is **never populated**, so both tools fail with "WIT directory not found".

## Root cause

The actual WIT content lives only in the system-level content-addressed store at `AstridHome::wit_dir()` = `~/.astrid/wit/{blake3hex}.wit`. Each installed capsule's `meta.json` carries `wit_files: HashMap<String, String>` mapping the original relative WIT filename (e.g. `broker.wit`, `deps/astrid-contracts/astrid-contracts.wit`) to its BLAKE3 hex hash. Install writes those content-addressed blobs via `astrid-capsule-install` (`content_address_wit` -> `home.wit_dir()`).

The capsule cannot reach that content store from capsule-space: its capability is `fs_read = ["home://"]`, and no VFS scheme resolves to `~/.astrid/wit/`. So the introspection tools see an empty (nonexistent) directory.

## Fix direction

At capsule-install time, after the WIT files are content-addressed into the store, also materialize a readable copy into the installing principal's `home://wit/` directory (`<principal_home>/wit/<basename>`), so the introspection tools can read them. This belongs in the shared install machinery (`astrid-capsule-install`) so both the CLI install path and the kernel `InstallCapsule` handler get it. Names are flattened to basename (`read_interface` rejects names containing `/`), dedup by basename, idempotent on re-install.

Verified manually: materializing `home://wit/` for the `default` principal makes both tools return real WIT content.

joshuajbouw:
## Linked Issue

Closes #906

## Summary

The system capsule's `list_interfaces` / `read_interface` introspection tools read WIT
interface contracts from `home://wit/`, but that directory was never populated. The actual
WIT content lives only in the BLAKE3-keyed content store at `~/.astrid/wit/`, which no VFS
scheme reaches from capsule-space (the system capsule's `fs_read` grant is `home://` only),
so both tools failed with "WIT directory not found".

This makes capsule install also materialize each capsule's content-addressed WIT into the
installing principal's `home://wit/` directory, so the introspection tools can read it. The
change lives in the shared `astrid-capsule-install` machinery, so both the CLI install path
and the kernel `InstallCapsule` handler get it.

## Changes

- `wit.rs`: new `materialize_wit_mirror(home, principal, wit_files)` — resolves
  `<principal_home>/wit/`, reads each blob back from the content store by hash, and writes a
  copy named by the WIT file's basename (`read_interface` rejects names containing `/`, so
  nested paths like `deps/astrid-contracts/astrid-contracts.wit` are flattened). Idempotent:
  byte-identical files are skipped, otherwise last-writer-wins via atomic temp + rename.
- `paths.rs`: new `install_principal()` — single source of truth for the install-target
  principal. `resolve_target_dir`, `resolve_env_path`, and the WIT mirror all resolve through
  it instead of each hardcoding `PrincipalId::default()`, so the capsule directory, env
  config, and interface mirror can never drift to different homes (per-principal installs
  become a one-line change here).
- `local.rs`: call the mirror after `write_meta`, passing `install_principal()`. Best-effort
  — a mirror failure logs a warning but does not roll back an otherwise-successful install
  (it degrades introspection visibility, not the install itself).
- `lib.rs`: re-export `content_address_wit` and `materialize_wit_mirror`.
- No host fn, VFS scheme, capability, or WIT/`.wit` change; content-store behavior untouched.

## Test Plan

- New `install_materializes_home_wit_mirror`: installs a capsule with a top-level and a
  nested WIT file; asserts both appear in `<home>/wit/` by basename with correct content and
  that the mirror is flat (no nested dirs leak through).
- New `install_wit_mirror_is_idempotent`: installs twice; asserts no error, identical
  content, and exactly one file (no temp-file leftovers).
- `cargo fmt --all -- --check`, `cargo clippy -p astrid-capsule-install --all-features
  --all-targets -- -D warnings`, `cargo test -p astrid-capsule-install` (10 passed), and
  `cargo check --workspace --all-features` all green.

## Notes / follow-ups

- Runs at install time, so already-installed capsules need a reinstall to backfill their WIT
  into `home://wit/`. A one-shot reconcile on daemon boot (or `astrid capsule reinstall
  --all`) would light up existing installs without manual reinstall — left as a follow-up.
- Alternative considered: a read-only `wit://` VFS scheme so the tool resolves the hash
  pointer and fetches the blob on demand with no per-principal duplication. Heavier (touches
  VFS resolution + a capability grant); deferred in favour of this zero-security-surface
  install-time mirror.


joshuajbouw:
Rebased onto current main (resolves the conflict) and addressed the review (e9c3e18):
- `materialize_wit_mirror` removes the partial temp file if `std::fs::write` fails (mirrors the existing rename-failure cleanup).
- Iterates `wit_files` in `rel_path`-sorted order so the basename-collision last-writer-wins outcome is deterministic, not HashMap-order-dependent.
- Reworded the `principal` doc to describe it as the mirror home target (workspace installs put the capsule dir under `<cwd>/.astrid`).
- Test `wit_mirror_dir` now resolves via `paths::install_principal()` instead of a hardcoded `PrincipalId::default()`.

Left the Windows rename-over-existing finding as-is: the daemon targets Linux/macOS (bwrap/seatbelt), so `cfg(windows)` overwrite handling would be untestable scope creep. Build + clippy + tests green.

Issue #856, opened 2026-06-06
Linux host_process sandbox still exposes arbitrary operator home and sibling project files read-only


joshuajbouw:
## Summary

The native `host_process` subprocess spawn path uses `SandboxCommand::wrap` (`core/crates/astrid-workspace/src/sandbox/mod.rs:63`), whose Linux arm is a hand-rolled bwrap invocation with no path-hiding mechanism:

```rust
// mod.rs:74-83 — SandboxCommand::wrap, Linux arm
bwrap
    .arg("--ro-bind").arg("/").arg("/")  // entire host fs readable RO
    .arg("--dev").arg("/dev")
    .arg("--proc").arg("/proc")
    .arg("--bind").arg(worktree_path).arg(worktree_path)
    .arg("--tmpfs").arg("/tmp")
    .arg("--unshare-all")
    .arg("--share-net")
    .arg("--die-with-parent");
```

`--ro-bind / /` grants recursive read access to the entire host filesystem, and there is no overlay to hide any subtree. A native subprocess spawned through this path can read `~/.astrid/` in full — audit log, KV store, secret store, capability store, and every principal's home directory — plus every sibling project and dotfile under the operator's home. The only scrubbing is `env_remove` of three vars (`ASTRID_HOME`, `ASTRID_SOCKET_PATH`, `ASTRID_SESSION_TOKEN`, `managed.rs:114-116`), which removes a path *hint* but not access to the well-known default `~/.astrid` location.

This is the same class of defect as #855 — a stale, divergent duplicate of the canonical sandbox builder missing a security-load-bearing rule — on the Linux arm instead of the macOS arm.

## Impact

`host_process` is the capability behind every native subprocess a capsule spawns (an LLM-supervisor capsule running `claude -p`, natively-spawned MCP stdio servers, build/test runners). On Linux today any such subprocess — including anything it re-execs, e.g. a shell — has read access to:

- `~/.astrid/` for **all** principals: secrets, KV, capability tokens, audit log, per-principal home dirs. A subprocess running on behalf of principal A can read principal B's secrets.
- Every sibling project and dotfile under the operator's home (`~/.ssh`, `~/dev/*`, etc.).

This is a multi-principal confidentiality break, and it defeats in-process tool-gating: a subprocess that can run a shell reads secrets regardless of which tools the agent layer allows, because the OS sandbox beneath it hides nothing.

[...]

PR #932, opened 2026-06-16
feat(auth): kernel-verified per-connection principal on publish-as + mask Astrid secret dirs in host_process sandbox

### LINKED ISSUE #931: Enforce kernel-verified per-connection principal on publish-as + mask Astrid secret dirs in the host_process sandbox
## Summary

Two coupled local multi-tenant hardening fixes, building on the per-connection ed25519 handshake already plumbed into the socket accept path.

1. **`publish-as` must stamp the kernel-verified connection principal, not the capsule-supplied name.** Implements the per-connection-binding mechanism proposed in #852 for connections that authenticate: the framed `tcp-stream.read` records the in-flight connection's verified principal (from the registry the handshake populates), and `publish-as` stamps THAT in place of the claimed name. A socket client that authenticates can no longer self-stamp a principal it has not proven. Unbound connections (the legacy peer-cred operator) fall back to the claim, so the zero-config operator CLI is unaffected.

2. **The `host_process` sandbox must hide `~/.astrid/{keys,secrets,var}` from spawned native processes.** Closes the Astrid-secret confidentiality break in #856: bwrap's `--ro-bind / /` exposed every principal's private keys, secret store, and the kernel state DB to any spawned native process. `wrap_with_injections` now overlays an empty tmpfs (Linux) / emits deny rules (macOS) on those three subpaths, keeping `run/` + `etc/` reachable so a spawned agent still reaches the daemon.

### Out of scope (follow-ups)
- fd-passing a pre-authenticated, principal-bound connection to daemon-spawned (`host_process`) agents + masking `run/` so they cannot dial a fresh unbound connection — completes #852's "every connection bound" and lets `publish-as` be deprecated.
- Broader agent ⊆ home read-isolation for sibling projects / `~/.ssh` (the rest of #856's impact).

Refs #852, #856.

joshuajbouw:
## Linked Issue

Closes #931. Refs #852 (implements its per-connection-binding mechanism), #856 (closes the Astrid-secret read in the host_process sandbox).

## Summary

Local multi-tenant hardening built on the per-connection ed25519 handshake already plumbed into the socket accept path: a socket client that authenticates can no longer claim a principal it has not proven, and a sandboxed native process can no longer read another principal's keys / secrets / kernel state. capsule-cli and the WIT are unchanged — the derivation is entirely host-side.

## Changes

- **`publish-as` enforcement (astrid-capsule).** New per-instance `HostState.ingress_principal`: the framed `tcp-stream.read` records the in-flight connection's verified principal from the `connection_principals` registry (uplink-gated; cleared on a non-data read), and `publish-as` stamps THAT over the capsule-supplied name when a binding exists, else falls back to the claim (the legacy peer-cred operator). `resolve_caller` reads the now-trustworthy principal. Path-2 (crypto-authenticated clients) enforces end-to-end.
- **`host_process` sandbox mask (astrid-workspace).** `wrap_with_injections` masks `~/.astrid/{keys,secrets,var}` — Linux empty `--tmpfs` overlay, macOS Seatbelt `with_hidden` deny — resolving the home via `AstridHome::resolve()` (fail-secure: a spawn whose home is unresolvable is refused). `run/` + `etc/` stay reachable so a spawned agent still reaches the daemon. The 400-line sandbox unit-test module moved to `sandbox/tests.rs` to keep `mod.rs` under the file-size cap.
- Earlier on the branch: per-principal ed25519 keypair minted at `agent create` + bootstrap; the two-frame challenge-response handshake; the per-connection registry.
- CHANGELOG `[Unreleased]` entries (Added: the auth enforcement; Security: the sandbox mask).

## Verification

- New tests: framed-read records/clears the ingress principal + non-uplink reads stay inert (`tcp_stream`); `publish-as` overrides a lying claim + falls back when unbound (`ipc`); the sandbox emits `--tmpfs` (Linux) / `deny file-read*` (macOS) for keys/secrets/var (`sandbox`).
- `cargo test -p astrid-capsule` (395 lib) + `-p astrid-workspace` green on macOS; the `astrid-workspace` sandbox suite green on Linux (`rust:1.95-slim` container).
- Empirically verified on Linux (bubblewrap in a container): `--ro-bind / /` reads a planted `~/.astrid/keys/*.key`; `--tmpfs` masks it while `run/system.token` stays readable; an inherited fd survives a full `~/.astrid` mask (the linchpin for the fd-pass follow-up).
- `cargo clippy --workspace --all-features` clean; `cargo fmt --all --check` clean. All commits GPG-signed.

## Checklist

- [x] Linked to an issue


joshuajbouw:
Addressed in 4a0c2b9 (after a self-review pass on top of this feedback):

1. **Private-key TOCTOU** (`mint_principal_keypair`, `mint_default_principal_keypair`) — both paths now create the key file atomically with `OpenOptions().mode(0o600)` *before* writing the secret bytes; the `write` → `set_permissions` chmod is gone, so the key is never briefly group/world-readable in the ~0755 keys dir. Both the bootstrap path and the `agent create` path are covered.

2. **`bwrap --tmpfs` on a non-existent mount point** — confirmed real and important: `secrets/` is created lazily (not by `AstridHome::ensure`), and `--tmpfs <missing>` under `--ro-bind / /` fails with `bwrap: Can't mkdir …: Read-only file system` (reproduced in an Ubuntu container), which would have aborted *every* `host_process` spawn on a fresh install. `sensitive_astrid_paths` now filters to existing paths — fail-safe, since an absent dir holds nothing to mask and rejoins the masked set once it is created. Regression test added and the mask tests made deterministic via a tempdir home.

Both verified: `cargo build`/`test`/`clippy --workspace --all-features`/`fmt` green (the two `socket::tests` failures are the known Unix-bind-under-sandbox environment issue, not these changes).

PR #934, opened 2026-06-16
feat(auth): stamp unauthenticated socket connections as no-capability anonymous (#933)

### LINKED ISSUE #933: Unbound socket connections still fall back to a claimed principal — stamp them no-capability `anonymous`
## Summary

The per-connection enforcement in #931 made the uplink's `publish-as` stamp the kernel-verified principal for a connection that authenticated. But it deliberately left the residual case — an **unbound** (unauthenticated) connection's `publish-as` still falls back to the principal it *claimed* — as a migration convenience.

That residual is the self-stamp itself: an unauthenticated local socket client can name (and act as) `default` (admin) or any principal, with no proof. #45/#852 calls for closing it.

## Proposal

Derive the principal entirely from the connection in `publish-as`:
- a kernel-verified binding → stamp that principal (as today);
- **no binding → stamp a reserved, no-capability `anonymous` identity** (no profile → capability checks fail closed; an unproven claim earns no privilege).

`anonymous` is reserved (`agent create` refuses it alongside `default`), so it can never be created or granted capabilities.

## Why it is safe (no live consumer relies on the fallback)

- The operator CLI and every agent authenticate — their ed25519 keypair is auto-minted, so they are bound, not unbound.
- The HTTP gateway publishes on the in-process event bus, bypassing the capsule-cli proxy / `publish-as` entirely (`astrid-gateway/src/state.rs`).
- The only thing the old fallback served was the self-stamp.

Kernel-internal, no WIT change. Completes #931's enforcement; refs #45/#852.

joshuajbouw:
## Linked Issue

Closes #933. Refs #852 (advances per-connection attribution; the `publish-as` deprecation it envisions remains future work).

## Summary

Completes #931's per-connection enforcement by closing the residual self-stamp it left. An **unbound** (unauthenticated) connection's `publish-as` no longer falls back to the principal it *claimed*. The principal is derived entirely from the connection — a kernel-verified binding stamps that principal; **no binding stamps a reserved no-capability `anonymous` identity** (no profile, so every capability check fails closed). Every socket connection is now either *authenticated* (acts as its proven principal) or *no-privilege*.

## Changes

- **`astrid-capsule` host `publish-as`:** derive the principal from the connection only — verified binding → that principal; unbound → `PrincipalId::anonymous()`. The capsule-supplied name is ignored entirely.
- **`astrid-core`:** `PrincipalId::anonymous()` — the reserved no-capability identity.
- **`astrid-kernel` admin:** reserve `anonymous` in `agent create` (extracted `reserved_principal_rejection`, beside the existing `default` reservation) so it can never be created or granted capabilities.
- **CHANGELOG** (Security entry) + corrected the now-stale "falls back to the supplied name" sentence in the #931 Added entry.

## Verification

- Updated/added tests: an unbound `publish-as` stamps `anonymous` (previously honoured the claim); `agent create` rejects `anonymous`.
- `cargo test -p astrid-capsule` (395 lib) + `-p astrid-kernel` + `-p astrid-core` green (the two `socket::tests` failures are the known Unix-bind-under-sandbox environment issue); `cargo clippy --workspace --all-features` clean; `cargo fmt --all --check` clean. Commit GPG-signed.
- Safety argument checked against the code: the operator and every agent authenticate (auto-minted keypair → bound); the HTTP gateway publishes on the in-process event bus, bypassing the capsule-cli `publish-as` path (`astrid-gateway/src/state.rs`); `BUILTIN_RESTRICTED` exists. The only thing the old fallback served was the self-stamp.

## Checklist

- [x] Linked to an issue


joshuajbouw:
Part of the local multi-tenant security floor epic — #935.

joshuajbouw:
Both addressed:

1. **`handlers.rs` reserved-name comparison** — already resolved before this review landed: the reserved-name logic moved into `astrid-core` as `PrincipalId::reserved_reason()`, which matches on `self.0.as_str()` (a string-slice comparison, no `PrincipalId`/`String` allocation). The handler now calls `principal.reserved_reason()`. (Your comment was on the pre-force-push revision.)

2. **`ipc.rs` double allocation** — fixed: the `Some` arm is now `verified.as_str().to_owned()` (single allocation) and the `None` arm is `PrincipalId::anonymous().into_inner()` (a new `into_inner()` moves the inner `String` out rather than `to_string()`-copying it), so each arm allocates exactly once.

PR #1007, opened 2026-06-23
feat(gateway): expose agent-loop readiness and fail fast on unconfigured chat

### LINKED ISSUE #963: Minimal capsule set for a working agent conversation is undocumented
_Found while integrating the **concierge** stack against AstridOS @ `cd43283` (LLM provider capsule + agent chat over the gateway). File/line refs are at that ref._

**Severity: major.**

`POST /api/agent/prompt` emits `ready` then nothing unless `react` + `session` + `identity` + an llm provider are all installed (`crates/astrid-gateway/src/routes/agent.rs` topic wiring; the dependency order is only visible in `toposort.rs` tests). The daemon hard-requires only `astrid-capsule-cli`. There's no manifest/endpoint that lists the minimal capsule set required for a functioning chat, so a fresh daemon that boots fine silently produces no replies.

**Ask:** document (or expose) the minimal capsule set for a working agent loop.

joshuajbouw:
## Linked Issue

Closes #963

## Summary

A fresh daemon boots clean but produces no agent replies unless the right capsules are loaded: `POST /api/agent/prompt` publishes `user.v1.prompt`, emits `ready`, then waits out the 5-minute stream timeout in silence if nothing subscribes the topic (or the LLM chain is incomplete). The daemon hard-requires only the socket uplink, and nothing reported what the agent loop was missing.

This exposes **agent-loop readiness**, computed name-agnostically from the loaded capsule manifests (never a hardcoded capsule list). A loop is ready when: (1) some loaded capsule `[subscribe]` matches `user.v1.prompt`, (2) some loaded capsule `[publish]` matches `agent.v1.response`, and (3) no required (non-optional) `[imports]` across the loaded set is unsatisfied. The kernel already computed import satisfaction at boot and discarded it; this consolidates that logic into one reusable, tested source of truth and surfaces it.

## Changes

- **Shared primitive** — `astrid-capsule/src/readiness.rs`: `agent_loop_readiness`, `topic_subscribers`/`topic_publishers` (route-layer subtree matcher), and `unsatisfied_required_imports`/`unsatisfied_optional_imports` (both apply the same cross-capsule self-exclusion, reusing `toposort::import_satisfied_by`). `validate_imports_exports` delegates to these, so the boot validator and the readiness report can never disagree.
- **Kernel API** — `KernelRequest::GetAgentReadiness` + `AgentLoopReadiness` DTO; handler reads the live registry. Gated `capsule:list` (mirrors `GetCapsuleMetadata`). Powers `GET /api/sys/readiness` (detailed, ops-facing) and the `astrid doctor` check.
- **`GET /api/sys/readiness`** — structured readiness for ops/admin (auth + `capsule:list`).
- **Multi-tenant fail-fast** — `POST /api/agent/prompt` probes readiness first; when not ready, it emits a single `error` SSE event naming the missing pieces and closes, instead of `ready` + a 5-minute silent timeout. The probe is **in-process** (`AgentReadinessProbe`, built by the kernel over the live registry and plumbed into the gateway), because agent-loop serviceability is global daemon health, not per-principal authorization — so the fail-fast fires for **every** authenticated prompt caller (single- and multi-tenant alike), with no capability check and no socket round-trip. Fails open when no probe is wired (standalone). The disclosure is narrowed: the prompt payload carries the missing WIT interface ids (public contract surface) but never capsule names or the inventory.
- **Boot WARN** — one concise, actionable line when the **loaded** registry (after load completes, not the pre-load discovered set) can't serve chat.
- **`astrid doctor`** — an "Agent loop readiness" PASS/WARN/FAIL check.
- **Security (folded in from the adversarial review): per-device capability scope on the gateway's `KernelClient` admin routes (Refs #999).** `KernelClient` (`astrid-uplink`) stamped only the caller's principal, never the authenticating device's `key_id`, so `GET /api/sys/status`, `GET /api/sys/readiness`, `POST /api/sys/capsules/reload`, and the `/api/capsules` list/get/install routes reached the kernel cap-gate as unattenuated full-principal authority — a device-scoped bearer that *denies* `capsule:list`/`reload`/`install` but whose principal *holds* it could still drive them. `KernelClient` now carries an optional `device_key_id` (`with_device_key_id`) and stamps it via a unit-tested message builder; all six call sites thread `CallerContext.device_key_id`, mirroring the `BusAdminClient` path. Full-authority bearers stamp `None` (unchanged).

All in-core; no capsule-repo change (the capsule-cli ingress allowlist is prefix-based on `astrid.v1.request.`).

## Verification

- `cargo clippy --workspace --all-features -- -D warnings` clean; `cargo fmt --all` applied; targeted `cargo test` green across the touched crates (capsule, kernel, gateway, uplink, core) plus the `gateway_e2e` integration test.
- **Regression tests** (each fails without its fix): readiness `not_ready_when_prompt_subscriber_absent` / `not_ready_when_required_import_unsatisfied` / `optional_import_satisfied_only_by_self_is_unsatisfied`; kernel `agent_readiness_probe_reflects_loaded_registry_without_capability`; gateway `unready_*`; `gateway_e2e` now drives `POST /api/agent/prompt` for a **non-admin** principal and asserts the in-process fail-fast (single `error` event, no `ready`, timeout-guarded); `KernelClient` `request_message_stamps_principal_and_device_key_id` / `..._omits_device_key_id_for_full_authority_caller`.
- Branch passed an adversarial multi-lens security review (capability/authz, kernel/IPC/concurrency, gateway/DoS, correctness, tests); confirmed findings were fixed or, for the device-scope gap, folded in.
- Manual reviewer steps: with only the socket uplink loaded, `GET /api/sys/readiness` returns `ready: false`, `astrid doctor` shows the readiness check FAIL, the daemon logs the boot WARN, and `POST /api/agent/prompt` returns an immediate `error` event instead of hanging — for any authenticated principal. Install a react + LLM provider + session + identity set and all flip to ready.

## Checklist

- [x] Linked to an issue
- [x] CHANGELOG.md updated (entries under `[Unreleased]`)


joshuajbouw:
Thanks both. Addressed in `c0f55ca`:

**Fixed**
- **Bound the readiness probe (connect + read).** `probe_agent_readiness` now wraps `KernelClient::connect` and the request in a 3s `PROBE_TIMEOUT` and fails open on expiry, so an unresponsive daemon can't make `POST /api/agent/prompt` hang — which would defeat the whole point of the fail-fast.
- **Boot readiness WARN now reflects the loaded set.** Moved `warn_agent_loop_readiness` to run after load completes, computing from the live registry rather than the pre-load discovered manifests — a manifest that fails to load (missing env, WASM error) is no longer mistaken for a working capability, keeping the boot signal aligned with `/api/sys/readiness`.
- **No more manifest cloning.** The readiness functions are now generic over `M: AsRef<CapsuleManifest>` (with a reflexive `AsRef` impl so tests keep passing `&[CapsuleManifest]`). The kernel handler, boot path, and `validate_imports_exports` pass `&[&CapsuleManifest]` borrowed straight from the registry — no per-prompt/per-boot clone.

**Deferred, with reasoning**
- **Probe uses the caller principal / `capsule:list` gating.** Correct catch. In single-tenant the caller is the default admin (`*`) so the fail-fast fires; a multi-tenant principal lacking `capsule:list` falls open to today's timeout behaviour — graceful degradation, not a regression. Widening readiness to every prompt-capable principal is a capability-model change (a "no-cap" introspection path or a privileged service-principal probe), which I'd rather make as a deliberate, separately-reviewed change than silently broaden here. Documented inline on `probe_agent_readiness`.

**Declined, with reasoning**
- **`Uuid::nil()` source_id in `astrid doctor`.** This mirrors the established CLI convention — `daemon_roundtrip` (GetStatus), `ps`, and `who` all build their request the same way; the connection already carries a fresh `SessionId` and the principal is bound at connect, so the message `source_id` isn't the attribution channel here. Deviating in just this one call site would make it inconsistent with its siblings. If nil-source-id is worth changing it's a codebase-wide convention cleanup, not a one-off in this PR.

Clippy `-D warnings` clean; readiness/gateway/kernel-router tests green.

joshuajbouw:
Second review pass — addressed in `2a4edd3`:

- **Nondeterministic readiness ordering** (`readiness.rs`): `agent_loop_readiness` now sorts `prompt_subscribers`, `response_publishers`, `loaded_capsules`, and `unsatisfied_required_imports`. The loaded set is `HashMap`-iterated, so raw order was unstable run-to-run — bad for an ops-facing API/CLI surface and a flaky-assertion risk for consumers. `MissingImport` gains `Ord`; a unit test pins the sorted output.
- **Hardcoded topics in `astrid doctor`**: `readiness_detail` now reuses `astrid_capsule::readiness::AGENT_PROMPT_TOPIC` / `AGENT_RESPONSE_TOPIC` instead of literal strings, so the CLI can't drift from the readiness computation.

Both low-severity. clippy `-D warnings` clean; readiness + core + cli tests green.

PR #1010, opened 2026-06-23
feat(cli): root-level shorthand for capsule CLI verbs

### LINKED ISSUE #1008: feat(cli): root-level shorthand for capsule CLI verbs (astrid <verb>)
# feat(cli): root-level shorthand for capsule CLI verbs (`astrid <verb>`)

> Repo: `unicity-astrid/astrid` (core) — crate `astrid-cli`.
> Parent: #961. Sibling issues referenced under **Scope boundaries** and **Dependencies & ordering**.

## Summary

`astrid capsule <verb> [args…]` already dispatches a capsule-contributed CLI verb
(`[[command]]` with `kind = "cli"`) to its providing capsule over IPC. This issue adds a
**root-level shorthand** so the same verb is reachable as `astrid <verb> [args…]` — e.g.
`astrid models list` instead of `astrid capsule models list`, and `astrid identity-export …`
instead of `astrid capsule identity-export …`.

`astrid capsule <verb>` stays as the canonical, unshadowable form; the root shorthand is a
convenience that resolves through the **exact same** daemon-backed resolution path. Built-in root
subcommands always win — a capsule can never shadow a built-in or intercept a mistyped built-in.

A **typo guard** preserves today's helpful behaviour: an unrecognised root token that is *close to*
a built-in (e.g. `astrid statsuss`) must emit a clap-style "did you mean `status`?" and exit
**without booting the daemon**, exactly as clap does today. Only a token that is *not* a near-miss
of any built-in is routed to the daemon for capsule-verb resolution.

This is a **generic mechanism**. It does not add the `models` verb (owned by the registry capsule,
sibling **3-registry**) and does not touch the HTTP gateway (sibling **2-gateway-api**). It
immediately benefits every existing capsule CLI verb (e.g. `identity-export`).

## Motivation

- **Ergonomics / scriptability parity with `gh`, `git`, `kubectl`.** Those tools let a plugin/extension
  verb be invoked at the root (`gh pr`, `git lfs`, `kubectl krew`) while keeping a canonical
  namespaced form. Astrid already commits to a "system verbs stay bare" philosophy
  (`crates/astrid-cli/src/main.rs:7-9`: "System verbs (`status`, `start`, `stop`, `ps`, `top`) stay
  as bare verbs for speed"). Capsule verbs deserve the same shorthand.
- **Programmatic integration.** Sibling **3-registry** exposes `astrid models {list,current,set,unset}`.
  Integrators will reach for `astrid models …` first; without this mechanism only the longer
  `astrid capsule models …` works.
- The resolution machinery already exists and is principal-scoped
  (`crates/astrid-cli/src/commands/capsule_verb.rs`); this issue only adds a second **entry point**
  into it plus a typo guard.

## Current behaviour (FILE:LINE)

All anchors verified at HEAD `3fb75db` on this checkout.

1. **Capsule verbs are reachable only under `astrid capsule …`.** The external-subcommand catch-all
   lives on the **`CapsuleCommands`** enum, not the root:
   - `crates/astrid-cli/src/cli.rs:386-395` — `#[command(external_subcommand)] External(Vec<String>)`
     on `CapsuleCommands`, with the doc-comment explaining built-ins shadow capsule verbs.
   - `crates/astrid-cli/src/dispatch.rs:260` — `CapsuleCommands::External(tokens) =>
     commands::capsule_verb::run_external(tokens).await`.
   - There is **no** `external_subcommand` on the root `Commands` enum today. Verified:
     `grep -rn "external_subcommand" crates/astrid-cli/src/` returns exactly one hit, `cli.rs:394`.

2. **The root `Commands` enum** (`crates/astrid-cli/src/cli.rs:92-282`) declares every built-in verb
   (`Chat`, `Run`, `Agent`, `Group`, `Caps`, `Quota`, `Invite`, `Keypair`, `PairDevice`, `Secret`,
   `Voucher`, `Trust`, `Audit`, `Budget`, `Session`, `Capsule`, `Mcp`, `Distro`, `Build`, `Init`,
   `Config`, `Wit`, `Gc`, `Start`, `Status`, `Stop`, `Restart`, `Logs`, `Ps`, `Top`, `Who`, `Doctor`,
   `Setup`, `Version`, `Completions`, `Update`). It is attached to `Cli` as
   `command: Option<Commands>` (`cli.rs:88-89`).

3. **An unknown root token is a clap error today, with suggestions, before any daemon contact.**
   `crates/astrid-cli/src/main.rs:49` calls `cli::Cli::parse()`. With no root `external_subcommand`,
   clap rejects an unrecognised first token (`astrid statuss`) by printing an "error: unrecognized
   subcommand … a similar subcommand exists: `status`" message and exiting the process *inside*
   `parse()` — `dispatch::dispatch` is never reached, the daemon is never started. **This behaviour
   must be preserved for built-in typos.**

4. **Resolution path to be reused** (`crates/astrid-cli/src/commands/capsule_verb.rs`):
   - `run_external(tokens: Vec<String>) -> Result<ExitCode>` (line 82) — first token is the verb,
     the rest are forwarded verbatim.
   - `match_verb(verb, &[CommandInfo]) -> VerbMatch` (line 61) — pure; filters to
     `kind == CommandKind::Cli && name == verb`; returns `One{provider,…}` / `None` / `Ambiguous`.
   - `resolve_commands()` (line 160) — `daemon::ensure_daemon("capsule")`, connects a `SocketClient`
     **bound to `crate::principal::current()`** (lines 171-174, RBAC scoping), sends
     `KernelRequest::GetCommands`, returns `Vec<CommandInfo>`.
   - `execute(provider, verb, &args)` (line 200) — publishes `cli.v1.command.run.<provider>`, awaits
     `cli.v1.command.result.<req_id>` with a 30s budget, renders, exits with the capsule's code.

5. **Reserved-name enforcement already exists** (defence in depth):
   - `crates/astrid-core/src/kernel_api.rs:205-208` — `RESERVED_CAPSULE_VERBS` = `["new","install",
     "update","list","remove","tree","deps","build","config","show","run","help"]`. **Note:** this
     list covers only `astrid capsule` built-in *sub*-verbs, **not** the root built-ins (`status`,
     `agent`, `mcp`, …). See **Threat model**.
   - `crates/astrid-capsule/src/discovery.rs:63-88` — `validate_cli_verb_name` rejects a `kind="cli"`
     command whose name is in `RESERVED_CAPSULE_VERBS` at manifest parse time (fail closed,
     `discovery.rs:82-86`).
   - `crates/astrid-cli/src/cli.rs:508-527` — `reserved_verbs_match_clap_subcommands` pins every
     `CapsuleCommands` clap variant name into `RESERVED_CAPSULE_VERBS`.

6. **Runtime introspection of built-in names is available.** `Cli::command()` (clap
   `CommandFactory`, already used at `crates/astrid-cli/src/commands/completions.rs:50`:
   `let mut cmd = Cli::command();`) yields a `clap::Command` whose `.get_subcommands()` enumerates the
   declared built-in subcommand names. `strsim` (clap's own suggestion crate) is already in
   `Cargo.lock`.

### 1. Add a root external-subcommand catch-all

In `crates/astrid-cli/src/cli.rs`, on the **root `Commands` enum** (the one ending at `cli.rs:282`
with `Update(UpdateArgs)`), add a final variant mirroring the existing `CapsuleCommands` one:

```rust
    /// Root shorthand for a capsule-provided CLI verb: `astrid <verb> [args…]`.
    ///
    /// Clap matches every declared variant above before falling through to
    /// this catch-all, so a capsule verb can never shadow a built-in. The
    /// canonical, unshadowable form remains `astrid capsule <verb>`.
    /// An unrecognised token that is a near-miss of a built-in is rejected
    /// with a "did you mean …?" hint *before* the daemon is contacted (see
    /// `dispatch`); only a non-near-miss token reaches capsule resolution.
    #[command(external_subcommand)]
    External(Vec<String>),
```

Constraints (clap 4.5, `derive`+`env`):
- The variant **must** be a single-field tuple over `Vec<String>` — clap's `external_subcommand`
  contract; identical to the existing `CapsuleCommands::External(Vec<String>)`.
- Global flags (`--verbose`, `--format`, `--principal`, `-p/--prompt`, …) are still parsed when they
  precede the verb; the first *non-flag* unrecognised token and everything after it land in the
  `Vec<String>`. Do **not** add `args_conflicts_with_subcommands`, `subcommand_required`, or a root
  positional — none exist today (`grep` confirms only `trailing_var_arg` at `cli.rs:383`, inside
  `CapsuleCommands::Run`), and adding them would break the existing no-subcommand interactive path
  (`dispatch.rs:121-129`) and the bare-prompt path (`dispatch.rs:62-84`).
- This intentionally **changes clap's default**: an unknown root token is no longer a clap parse
  error — it becomes `Commands::External([token, …])`. Clap's built-in "did you mean" engine no
  longer fires for root tokens, which is exactly why the typo guard below is mandatory.

### 2. Add a pure typo-guard function

New module `crates/astrid-cli/src/commands/verb_suggest.rs` (or a `pub(crate)` fn beside
`capsule_verb`), with **no daemon dependency**:

```rust
/// Closest built-in subcommand name to `token`, if one is "close enough"
/// to be a probable typo. Pure over `(token, builtin_names)` so it is
/// unit-testable without a daemon or clap context.
///
/// Returns `Some(builtin)` when the nearest built-in is within an edit
/// distance the user almost certainly meant (Damerau/Levenshtein ≤ a
/// length-scaled threshold; mirror clap's own heuristic — `strsim` is
/// already a transitive dep). Returns `None` when nothing is close, so
/// the caller routes the token to capsule-verb resolution.
pub(crate) fn nearest_builtin<'a>(token: &str, builtin_names: &[&'a str]) -> Option<&'a str>;
```

Implementation notes:
- Threshold must err **toward routing to the daemon** (return `None`) when uncertain, so that a real
  capsule verb that merely *rhymes* with a built-in is not stolen by the guard. A conservative rule:
  treat as a typo only when the best edit distance is `≤ max(2, len/3)` **and** the token is not an
  exact match of any built-in (an exact match can't reach here — clap consumed it). Pick the
  lowest-distance built-in; break ties alphabetically for determinism.
- The built-in name list is supplied by the caller, derived once from `Cli::command()` (see step 3),
  so the function never re-derives clap state and stays trivially testable with a literal slice.

### 3. Route the root `External` in dispatch with the typo guard in front

In `crates/astrid-cli/src/dispatch.rs`, add a `Some(Commands::External(tokens))` arm to
`dispatch_subcommand` (the `match command` at `dispatch.rs:111`). Before contacting the daemon:

```rust
        Some(Commands::External(tokens)) => {
            let verb = tokens.first().map(String::as_str).unwrap_or("");
            let builtins = builtin_subcommand_names(); // from Cli::command().get_subcommands()
            if let Some(suggestion) = commands::verb_suggest::nearest_builtin(verb, &builtins) {
                eprintln!(
                    "{}",
                    theme::Theme::error(&format!(
                        "unrecognized subcommand '{verb}'\n\n\
                         \tDid you mean '{suggestion}'?"
                    ))
                );
                // Exit WITHOUT booting the daemon — mirrors today's clap error.
                return Ok(ExitCode::from(2));
            }
            // Not a near-miss → genuine capsule-verb shorthand.
            commands::capsule_verb::run_external(tokens).await
        },
```

- `builtin_subcommand_names()` is a small helper that calls `Cli::command()` and collects
  `.get_subcommands().map(|s| s.get_name())`. It must **exclude** the `External` catch-all itself
  (clap reports it with an empty/placeholder name — filter empties). Do this once; it allocates a
  `Vec<String>`; borrow `&[&str]` into `nearest_builtin`.
- Exit code **2** for the typo path matches clap's `ErrorKind` convention for usage errors (clap
  exits `2` on `InvalidSubcommand`), keeping script behaviour stable vs. today.
- The happy path delegates to the **unchanged** `run_external` — no new IPC, principal scoping, or
  rendering logic. `run_external` already binds `principal::current()` (`capsule_verb.rs:171-174`),
  so the shorthand inherits the same RBAC scoping as `astrid capsule <verb>`.

### 4. Precedence & co-existence (no new code, assert in tests)

- `astrid capsule <verb>` (`CapsuleCommands::External`, `cli.rs:394`) is **unchanged** and remains the
  canonical form.
- Built-in root verbs win automatically: clap matches declared `Commands` variants before the
  `external_subcommand` catch-all (same guarantee the existing `CapsuleCommands` doc-comment at
  `cli.rs:386-393` relies on). `astrid status`, `astrid mcp serve`, etc. continue to dispatch to
  their built-ins; they never reach `run_external`.

## Acceptance criteria

- [ ] `astrid <verb> [args…]` resolves a `kind="cli"` capsule verb through the same
      `capsule_verb::run_external` path as `astrid capsule <verb>` (same provider resolution, same
      `cli.v1.command.run.<provider>` dispatch, same principal scoping, same exit-code mapping).
- [ ] `astrid capsule <verb>` continues to work identically; both forms co-exist.
- [ ] Every built-in root verb still dispatches to its built-in (verified for at least
      `status`, `mcp serve`, `capsule list`, `version`) and never reaches `run_external`.
- [ ] An unrecognised root token that is a near-miss of a built-in (`astrid statuss`,
      `astrid agnet`) prints a "did you mean '<builtin>'?" message and exits **2** **without** booting
      the daemon (no socket connection, no `ensure_daemon`).
- [ ] An unrecognised root token that is **not** a near-miss (`astrid identity-export`,
      `astrid models`) is routed to daemon-backed capsule resolution (and, if no capsule provides it,
      produces `run_external`'s existing "Unknown capsule command" + available-list output, exit 1 —
      `capsule_verb.rs:100-107`).
- [ ] `nearest_builtin` is a pure function over `(token, &[&str])` with no daemon/clap dependency.
- [ ] CHANGELOG.md updated under `[Unreleased]`.
- [ ] `cargo test --workspace` passes; no new clippy warnings under the crate's `#![deny(clippy::all)]`.

## Tests (named regression tests)

Each test fails on `main` before the change.

In `crates/astrid-cli/src/commands/verb_suggest.rs` (`#[cfg(test)] mod tests`):

- `nearest_builtin_flags_obvious_builtin_typo` — `nearest_builtin("statuss", &BUILTINS)` →
  `Some("status")`; `nearest_builtin("agnet", &BUILTINS)` → `Some("agent")`. Fails today (function
  does not exist).
- `nearest_builtin_passes_through_real_capsule_verb` — `nearest_builtin("identity-export", &BUILTINS)`
  → `None` and `nearest_builtin("models", &BUILTINS)` → `None` (neither is within the typo threshold
  of any built-in), so the dispatcher routes them to the daemon instead of hijacking them.
- `nearest_builtin_exact_builtin_is_not_a_suggestion` — `nearest_builtin("status", &BUILTINS)` →
  `None` (an exact built-in never reaches the guard via clap; the function must not treat distance-0
  as a typo). Guards against accidentally short-circuiting a name clap already handled.
- `nearest_builtin_is_deterministic_on_ties` — a token equidistant from two built-ins returns the
  alphabetically-first, so script output is stable.

In `crates/astrid-cli/src/cli.rs` (`#[cfg(test)] mod tests`, alongside the existing
`reserved_verbs_match_clap_subcommands`):

- `root_external_subcommand_captures_unknown_verb` — using `Cli::try_parse_from(["astrid",
  "frobnicate", "--flag", "x"])`, assert the parse yields `Some(Commands::External(v))` with
  `v == ["frobnicate", "--flag", "x"]`. Fails today (no root `external_subcommand`; clap errors).
- `root_builtin_wins_over_external` — `Cli::try_parse_from(["astrid", "status"])` parses to
  `Some(Commands::Status)`, **not** `External`. Pins precedence so a future refactor can't let the
  catch-all swallow a built-in.
- `builtin_subcommand_names_excludes_external_placeholder` — assert the names harvested from
  `Cli::command().get_subcommands()` (the list fed to `nearest_builtin`) contains `status`/`agent`
  but contains no empty string and no `External`-placeholder entry, so the guard never "suggests" the
  catch-all itself.

> If the typo-guard wiring inside `dispatch_subcommand` is hard to test without a daemon, the
> load-bearing logic is already extracted into the pure `nearest_builtin` and the clap parse tests
> above; the dispatch arm is a thin call into them. No daemon-backed integration test is required for
> this issue.

## Threat model / security

The shorthand is a second **entry point** into an already principal-scoped resolution path; it adds
no new authority. Specific concerns:

1. **A capsule must not shadow a built-in root verb.** Clap matches declared `Commands` variants
   before the `external_subcommand` catch-all, so `astrid status` can never reach a capsule even if a
   capsule declares a `status` CLI verb. **Gap to note in the PR:** `RESERVED_CAPSULE_VERBS`
   (`kernel_api.rs:205-208`) only lists `astrid capsule` *sub*-verbs; it does **not** list the root
   built-ins (`status`, `agent`, `mcp`, …). A capsule *can* therefore declare a `kind="cli"` verb
   named `status`. That verb stays invisible at the **root** (clap's built-in `Status` wins) but is
   still reachable as `astrid capsule status` and `astrid capsule run <provider> status`. This is the
   pre-existing, documented contract — the root shorthand does **not** regress it (the built-in still
   wins at the root). **This issue must not silently expand `RESERVED_CAPSULE_VERBS` to root names**;
   doing so would break existing capsules and belongs in its own change if desired. The residual,
   accepted tradeoff: a **future** core built-in added to `Commands` will start shadowing a capsule's
   root shorthand of the same name. This is acceptable because `astrid capsule <verb>` remains the
   stable, unshadowable fallback — call this out explicitly in the PR description.

2. **A capsule must not intercept a mistyped built-in.** The typo guard runs **before** any daemon
   contact: a token close to a built-in is rejected with a suggestion and never published to the bus,
   so a capsule cannot register a near-miss verb (`statuss`) to harvest fat-fingered admin commands.
   The guard's bias-toward-`None` only widens what reaches the daemon for *non*-near-miss tokens; it
   never narrows the built-in protection.

3. **No privilege change on the resolution path.** `run_external` already binds the connection to
   `principal::current()` (`capsule_verb.rs:171-174`) and stamps the principal on the `GetCommands`
   and `run` messages, so the shorthand enumerates and dispatches capsule verbs under the invoking
   identity — never the `default`/admin principal a nil/unstamped message would fall back to. The
   shorthand reuses this verbatim; it must not introduce a separate, unbound `SocketClient`.

4. **Daemon-boot avoidance for typos is a DoS/UX property, not just cosmetics.** Keeping the
   "did you mean" path daemon-free means a typo storm (or a tab-completion misfire) cannot
   repeatedly auto-start the daemon (`daemon::ensure_daemon`, `capsule_verb.rs:162`).

## Scope boundaries

In scope (this issue, `core` / `crates/astrid-cli` only):
- Root `external_subcommand` catch-all on `Commands` (`cli.rs`).
- Pure `nearest_builtin` typo guard + its wiring in `dispatch_subcommand` (`dispatch.rs`).
- Tests above; CHANGELOG entry.

Explicitly **not** in this issue:
- The `models` verb and any registry-side selection logic — **out of scope, owned by sibling
  `3-registry`** (`unicity-astrid/capsule-registry`). This issue is a generic mechanism; it ships and
  is tested against existing capsule verbs (e.g. `identity-export`) with no `models` dependency.
- The HTTP / gateway mirror of capsule-verb invocation — **out of scope, owned by sibling
  `2-gateway-api`** (`unicity-astrid/astrid`, gateway crate).
- Any change to `RESERVED_CAPSULE_VERBS` or to whether root built-in names are reserved against
  capsules — **out of scope** (see Threat model #1); the root built-in already wins structurally.
- The `react`-side model plumbing and registry contract honouring — **out of scope, parent #961 /
  sibling registry issues.**

## Dependencies & ordering

- **Depends on:** nothing. This is a self-contained `astrid-cli` change and can land independently of
  the registry and gateway siblings. It does not modify any WIT, IPC contract, or kernel code — no
  RFC required.
- **Unblocks / enables:** sibling `3-registry`'s `astrid models …` becomes reachable as the root
  shorthand the moment the registry ships its `kind="cli"` `models` verb. Any current/future capsule
  CLI verb (e.g. `identity-export`) gains the shorthand for free.
- **Parent:** #961.

joshuajbouw:
## Linked Issue

Closes #1008

## Summary

Adds a root-level shorthand so a capsule-contributed `kind = "cli"` verb is reachable as `astrid <verb> [args…]` (e.g. `astrid identity-export …`, and `astrid models …` once the registry ships it) — the same convenience `gh`/`git`/`kubectl` give their plugin verbs. `astrid capsule <verb>` stays the canonical, unshadowable form. The shorthand is a second entry point into the **unchanged**, principal-scoped resolution path (`commands::capsule_verb::run_external`); it adds no new authority, no new IPC, and no WIT change.

## Changes

- **Root `external_subcommand` catch-all** (`cli.rs`): a final `Commands::External(Vec<String>)` variant mirroring the existing `CapsuleCommands::External`. Clap matches every declared built-in variant before the catch-all, so a capsule can never shadow a built-in (`status`, `mcp serve`, `version`, …) at the root.
- **Pure typo guard** (`commands/verb_suggest.rs`): `nearest_builtin(token, &[&str])` — Levenshtein `≤ max(2, len/3)`, exact matches and empty tokens excluded, ties broken alphabetically for deterministic output. No daemon/clap dependency.
- **Dispatch wiring** (`dispatch.rs`): the root `External` arm runs the guard *before* any daemon contact. A near-miss of a built-in (`astrid statuss`) prints "did you mean 'status'?" and exits **2 without booting the daemon**; a non-near-miss (`astrid identity-export`) is forwarded to `run_external`. Built-in names are harvested once from `Cli::command()`, filtering the empty `External` placeholder.

## Security notes

- A capsule cannot shadow a built-in root verb — clap structurally prefers declared variants; this PR does **not** regress that and deliberately does **not** expand `RESERVED_CAPSULE_VERBS` to root names (that would break existing capsules and belongs in its own change). The root built-in already wins structurally.
- A capsule cannot intercept a mistyped built-in — the typo guard runs before any bus publish, so a near-miss verb is never routed to the daemon and exits without booting it (a DoS/UX property, not just cosmetic).
- No privilege change: the shorthand reuses `run_external`, which binds `principal::current()` and stamps it on `GetCommands`/`run` — same RBAC scoping as `astrid capsule <verb>`.
- **Accepted residual tradeoff:** a *future* core built-in added to `Commands` will begin shadowing a capsule's root shorthand of the same name. Acceptable because `astrid capsule <verb>` remains the stable, unshadowable fallback.

## Verification

- Code: `cargo test --workspace` passes (the only failures observed locally are two gateway TLS integration tests that bind ephemeral TCP ports, which the local sandbox blocks — they pass with the sandbox disabled and are unrelated to this change). No new clippy warnings under `--workspace --all-features -D warnings`.
- New regression tests (each fails on `main`): `verb_suggest` unit tests (`nearest_builtin_flags_obvious_builtin_typo`, `_passes_through_real_capsule_verb`, `_exact_builtin_is_not_a_suggestion`, `_is_deterministic_on_ties`, plus an empty-token guard) and clap parse tests in `cli.rs` (`root_external_subcommand_captures_unknown_verb`, `root_builtin_wins_over_external`, `builtin_subcommand_names_excludes_external_placeholder`).

## Review fixes

Addressed a review finding that the `BUILTINS` slice backing the `nearest_builtin` typo-guard unit tests had drifted from the real root subcommand set: it omitted `pair-device` and the `self-update` alias. The slice now lists both, and a new `builtins_slice_matches_clap_subcommands` test diffs it against the live `Cli::command()` tree (primary names plus invocable aliases) so it can no longer silently drift when a root subcommand is added or renamed. This is a test-fidelity fix only - production still harvests builtins dynamically from `Cli::command()`.

## Rebase

Rebased onto current `origin/main` (now at the gateway agent-loop-readiness feature, #1007). The only conflict was `CHANGELOG.md`, where both `[Unreleased]` sections added entries; resolved by keeping both (main's readiness entry and this branch's CLI-shorthand entry). The CLI code does not overlap with main's changes. `build` / `clippy --workspace --all-features -D warnings` / `test --workspace` all green afterward (modulo the two known sandbox-only ephemeral-port TLS tests).

## Checklist

- [x] Linked to an issue


joshuajbouw:
Folded in the triaged review findings:

- **Perf (Copilot):** `nearest_builtin` re-collected the typed token into a `Vec<char>` on every loop iteration (inside `levenshtein`). The token's chars are now collected once before the loop and reused across every built-in comparison via a char-slice `levenshtein_chars`. Behaviour is identical — all six `verb_suggest` unit tests (tie-break determinism, exact-match, empty-token, pass-through) pass unchanged.
- **Doc (Copilot):** softened the `builtin_subcommand_names` doc that said "harvested once" while rebuilding the clap tree per call. Took the cheap correct option: a CLI process dispatches a single command and exits, so the lone caller invokes it at most once per process — the doc now states it is built fresh and no caching is warranted, rather than adding a `OnceLock` that would never pay off in a single-shot process.

- **False positive confirmed (Copilot, cli.rs:~553):** the claim that the `Commands::External` (`Vec<String>`) vs `vec!["frobnicate", ...]` (`Vec<&str>`) assertion "won't compile" is incorrect. `Vec<String>: PartialEq<Vec<&str>>` holds via the stdlib blanket `impl<A: PartialEq<B>, B> PartialEq<Vec<B>> for Vec<A>` combined with `String: PartialEq<&str>`. `root_external_subcommand_captures_unknown_verb` compiles and passes — no change made.

Full `-p astrid` suite green (255 passed).

joshuajbouw:
Folded the second-round re-review findings on this branch:

**Folded (correctness):** The production typo guard `builtin_subcommand_names()` in `crates/astrid-cli/src/dispatch.rs` harvested only primary subcommand names, omitting invocable aliases like `self-update` (an alias of `update`). So a mistyped alias (`self-updte`) matched no near-miss and got routed to the daemon as a capsule verb instead of suggesting "did you mean 'self-update'?". It now harvests every alias via `get_all_aliases()` alongside the primary name and de-duplicates through a `BTreeSet`. Added three regression tests against the real production harvest: one asserts `self-update` is present, one asserts the empty catch-all placeholder is filtered, and an end-to-end one asserts `nearest_builtin("self-updte")` returns `Some("self-update")` without daemon contact. I verified the two alias-targeting tests fail when the harvest drops aliases, so they're genuine regression guards.

**Folded (perf):** `verb_suggest::nearest_builtin` now reuses a single `Vec<char>` buffer across the per-name levenshtein collect (`.clear()` + `extend`) instead of allocating a fresh vector per built-in.

**Dismissed:** The flagged assertion in `cli.rs` comparing `Commands::External(Vec<String>)` to a `vec!["frobnicate", ...]` of `&str` is not a bug — `Vec<String>: PartialEq<Vec<&str>>` makes it compile and pass. To stop the recurring re-flag I rewrote the assertion to compare owned `String`s explicitly; it's a readability change, not a fix.

Build green, clippy clean on all touched files, targeted tests pass. (The branch already had pre-existing `--all-targets` clippy lints in unrelated test modules — `tui/state.rs`, `daemon_control.rs`, `logs.rs`, `quota.rs`, `distro/validate.rs` — from a newer clippy toolchain; those are out of scope for this fold and untouched.)

joshuajbouw:
Folded the two perf nits from the bot review:

- `nearest_builtin` is now generic over `S: AsRef<str>`, so `dispatch_root_shorthand` passes the harvested `Vec<String>` of builtin names directly — dropping the intermediate `Vec<&str>` re-collection on the hot path. The returned `&str` borrows from the input slice; the existing `&str`-literal test slices still satisfy `AsRef<str>` unchanged.
- The Levenshtein rolling rows are now sized to `min(a, b) + 1` by driving the DP with the shorter string. Levenshtein is symmetric, so swapping which string indexes the rows leaves every distance identical, and the single-buffer reuse from 1f34381 is preserved. Added a `levenshtein_is_symmetric` regression test.

Both changes are behaviour-preserving — suggestions and edit distances are unchanged. `cargo build`, `cargo clippy --workspace --all-features -D warnings`, and `cargo test -p astrid` (259 passed, incl. the new symmetry test) are all green.

joshuajbouw:
Rebased onto current origin/main to resolve a CHANGELOG `[Unreleased]` conflict — kept both entries (main's readiness entry and this branch's CLI-shorthand entry); the CLI code did not overlap. Build / clippy --all-features / test all green afterward (new tip 2bb5be1).

PR #1011, opened 2026-06-23
feat(gateway): HTTP API to list and bind the active LLM model per principal

### LINKED ISSUE #1009: feat(gateway): HTTP API to list and bind the active LLM model per principal
# feat(gateway): HTTP API to list and bind the active LLM model per principal

**Repo:** `unicity-astrid/astrid` (core) — crate `crates/astrid-gateway` only
**Parent:** #961

## Summary

The HTTP gateway exposes no way to discover or bind a principal's active LLM
provider/model. The route table in `crates/astrid-gateway/src/routes/mod.rs`
(`build`, lines 31–178) wires principals, caps, quotas, groups, invites,
capsules, env, audit, agent-prompt and system endpoints — but nothing for the
model registry. There is no `/api/models*` entry in the `authed` block
(mod.rs:57–140).

This issue adds three principal-scoped routes that wrap the **existing**
registry IPC contract:

- `GET  /api/models`         — list available provider-entries for the caller
- `GET  /api/models/active`  — the caller's currently bound model (or `null`)
- `PUT  /api/models/active`  — bind by `{ "id": "<capsule>:<model>" }`

The gateway is a thin transport: it publishes the registry request IPC, awaits
the registry's reply on a **principal-scoped routed subscription**, and surfaces
the registry's own success/error verbatim. It does **not** resolve ids,
enumerate models, or persist selection — that is the registry capsule's job
(parent #961, sibling registry issue).

A per-request model override on `POST /api/agent/prompt` is **explicitly NOT in
this issue's scope** — see _Scope boundaries_ and _Open questions_ for why and
what would have to land first.

## Motivation

The parent issue (#961) is driven by *programmatic* integration: a script or
remote dashboard must be able to read the available models for a principal and
bind one without driving the TUI's interactive `/models` slash command. Today
the only selection surface is that TUI picker. A web/dashboard/automation client
talking to the gateway has no route to call. Machine-readable JSON is therefore
mandatory (these handlers return JSON natively; no `--json` flag needed as in the
CLI sibling).

The active-model binding is **per principal** — registry state (provider list +
active model id) lives in the capsule's KV, which the kernel scopes by invocation
principal (documented at `capsules/astrid-capsule-registry/src/lib.rs:192–207`).
The gateway must preserve that isolation end-to-end: a caller may only read and
set *its own* active model.

## Current behaviour (FILE:LINE)

- **No model routes.** `crates/astrid-gateway/src/routes/mod.rs` `build()` —
  the `authed` router block (mod.rs:57–140) registers every authenticated route;
  there is no `/api/models*` entry. The agent route sits at mod.rs:130.
- **Registry IPC already exists** and is what these handlers wrap
  (`capsules/astrid-capsule-registry/src/lib.rs`):
  - Request topics handled in `dispatch_registry_messages` (lib.rs:326–334):
    - `registry.v1.get_providers`   → responds `registry.v1.response.get_providers`
      with a JSON **array** of `provider-entry` (lib.rs:177–190).
    - `registry.v1.get_active_model`→ responds `registry.v1.response.get_active_model`
      with a single `provider-entry` **or `null`** (lib.rs:208–238).
    - `registry.v1.set_active_model`→ responds `registry.v1.response.set_active_model`
      with `{ "status": "ok", "active_model": <provider-entry> }` on success or
      `{ "error": "unknown model: <id>" }` / `{ "error": "missing model_id" }` on
      failure (lib.rs:240–283). Request payload shape: `{ "model_id": "<id>" }`
      (also accepted nested under `data.model_id`, lib.rs:242–246).
  - **Resolution of `<capsule>:<model>` / bare ids and ambiguity is the
    registry's job** (parent #961, design point B / sibling registry issue). The
    gateway forwards the raw `id` and relays the registry's reply unchanged.
- **The two gateway dispatch patterns** are both present in this crate; use the
  bus one (registry is a capsule, not a kernel admin op):
  - In-process **event bus** publish + routed subscribe + await reply — see
    `agent.rs:133–187` (`state.event_bus`, `subscribe_topic_routed`, `bus.publish`).
  - Kernel admin `KernelClient::connect(caller.principal)` — see
    `system.rs:35` (and the `#[utoipa::path]` annotation at system.rs:20). **Not**
    applicable here (registry is capsule IPC, not a kernel admin op).
- **Principal extraction** is uniform: `caller_from(&req)` returns the verified
  `CallerContext` from request extensions. agent.rs imports it from
  `crate::routes::principals::caller_from`, which is a `pub(crate) fn` at
  `crates/astrid-gateway/src/routes/principals.rs:541` (a distinct definition
  from the `pub fn caller_from` in `auth.rs:292`; use the `routes::principals`
  path, matching agent.rs:64). `CallerContext.principal` is the only trusted
  identity (`auth.rs:69–85`).
- **Body parsing** helper: `read_json_body::<T>(req)` caps at 64 KiB
  (`principals.rs:567`).
- **Error surface:** `GatewayError` variants `Unauthorized` (error.rs:47),
  `Forbidden { reason }` (error.rs:51), `BadRequest(String)` (error.rs:57),
  `NotFound` (error.rs:60), `Internal(#[from] anyhow::Error)` (error.rs:72);
  `IntoResponse` mapping at error.rs:78–112.
- **OpenAPI doc** is generated; handlers + schemas must be registered in
  `crates/astrid-gateway/src/openapi.rs` under `paths(...)` (line 64) and
  `components(schemas(...))` (lines 122–123). Example: `routes::agent::post_prompt`
  at openapi.rs:109, `PromptRequest` at openapi.rs:166.

## Proposed change (implementer-ready)

Add `crates/astrid-gateway/src/routes/models.rs`, declare `pub mod models;` in
`routes/mod.rs`, and register the three routes inside the `authed` router block
in `build()`.

### 1. New module `crates/astrid-gateway/src/routes/models.rs`

All three handlers share one helper that does the bus round-trip,
**scoped to the authenticated principal**. The isolation seam is
`EventBus::subscribe_topic_routed_scoped(.., scope)`
(`crates/astrid-events/src/bus.rs:317`): a scoped route admits only events whose
publisher principal equals `scope` and drops foreign-principal events at enqueue
(`accepts()` at `crates/astrid-events/src/route/entry.rs:168–169`, enforced as a
top-of-`push_with_eviction` guard at `route/entry.rs:187–189`).

**Why the registry reply is stamped with the requester's principal (verified).**
The scoped subscription only works if the registry's reply to a request stamped
with principal `P` carries `message.principal == P` (and not the registry
capsule's own load-time identity). That holds, traced through the host:

1. When the kernel dispatches an inbound IPC message to the registry's
   interceptor, `WasmEngine::invoke_interceptor` sets
   `state.caller_context = caller.cloned()`
   (`crates/astrid-capsule/src/engine/wasm/mod.rs:2106`), where `caller` is the
   inbound message — stamped with `P`.
2. The registry replies via the SDK `publish`, whose host implementation resolves
   the outgoing principal from `caller_context.principal`, **falling back to the
   capsule's own `self.principal` only when no caller is in scope**
   (`crates/astrid-capsule/src/engine/wasm/host/ipc.rs:291–295`). With a caller in
   scope (`P`), the reply is stamped `P`.
3. `publish_inner` then sets `message.principal = P` via `.with_principal(...)`
   (`ipc.rs:272–273`), and `effective_principal()` (the budget/audit principal)
   likewise prefers `caller_context` over the owner
   (`crates/astrid-capsule/src/engine/wasm/host_state_effective.rs:90–96`).

Therefore `registry.v1.response.*` for a request stamped `P` carries
`message.principal == P`, and a subscription scoped to `Some(Some(P))` receives
only `P`'s reply. (The registry uses own-principal `publish` for its replies, not
`publish_as`, so no device-scope/forward path is involved.)

Helper (shape; names illustrative):

```rust
/// Publish `request_topic` (stamped with the caller's principal) and await
/// the single reply on `response_topic`, scoped to the caller so no other
/// principal's registry reply can be observed. Mirrors the subscribe-first /
/// publish-second ordering and timeout discipline of agent.rs.
async fn registry_round_trip(
    state: &GatewayState,
    caller: &CallerContext,
    request_topic: &'static str,
    response_topic: &'static str,
    payload: serde_json::Value,         // RawJson body; {} for the GETs
) -> GatewayResult<serde_json::Value>   // the reply payload as JSON
```

Implementation requirements:

1. Resolve `bus = state.event_bus.clone()` — if `None`, return
   `GatewayError::Internal` with the same message style as `agent.rs:139–143`
   ("gateway is not wired to a live event bus").
2. **Subscribe first, then publish** (agent.rs:151–175 documents the race). Use
   `bus.subscribe_topic_routed_scoped(Uuid::new_v4(), response_topic, "gateway",
   "gateway::models", Some(Some(caller.principal.to_string())))`. **Note the
   double `Some`:** the parameter type is
   `scope: Option<PrincipalKey>` where `PrincipalKey = Option<String>`
   (`crates/astrid-events/src/route/entry.rs:62`). The outer `Some` means "this
   route is scoped"; the inner `Some(principal)` is the principal it is scoped to
   — exactly the form the host uses at
   `crates/astrid-capsule/src/engine/wasm/host/ipc.rs:411`
   (`Some(Some(self.principal.to_string()))`). Passing a bare
   `Some(caller.principal.to_string())` is a type error and will not compile; a
   single `None` would disable the security scope. The fresh per-call UUID
   isolates this connection's route (agent.rs:159–167); the inner `Some(principal)`
   is the security boundary.
3. Publish an `AstridEvent::Ipc` with `IpcMessage::new(request_topic,
   IpcPayload::RawJson(payload), Uuid::nil()).with_principal(caller.principal
   .to_string())` (mirror agent.rs:177–187). The principal stamp is mandatory —
   it both routes the request to the caller's registry KV scope and is the value
   the reply is matched against.
4. Await the reply with a bounded timeout (use a `const` like
   `REGISTRY_TIMEOUT: Duration = Duration::from_secs(10)`, matching the 10s budget
   used for daemon reads in `capsule_verb.rs:186`). On timeout return
   `GatewayError::Internal` ("registry did not respond"). Extract the reply
   payload from the matched `AstridEvent::Ipc { message, .. }` exactly as
   `agent.rs:280–296` does (`IpcPayload::RawJson(v)` → `v`, else `to_value`).

Handlers:

- **`GET /api/models` → `list_models`**
  - `caller = caller_from(&req)?.clone();`
  - `registry_round_trip(.., "registry.v1.get_providers",
    "registry.v1.response.get_providers", json!({}))`
  - The reply is a JSON array of provider-entries. Return `Json(reply)` (200).
    Pass the array through unchanged; do not reshape.

- **`GET /api/models/active` → `get_active_model`**
  - `registry_round_trip(.., "registry.v1.get_active_model",
    "registry.v1.response.get_active_model", json!({}))`
  - Reply is a single provider-entry **or JSON `null`**. Return `Json(reply)`
    (200) including the `null` case — `null` means "no model bound", not an error.

- **`PUT /api/models/active` → `set_active_model`**
  - Body: `SetActiveModelRequest { id: String }` (deserialized via
    `read_json_body`). Reject an empty/whitespace `id` with
    `GatewayError::BadRequest("id must not be empty")` **before** contacting the bus.
  - Publish `registry.v1.set_active_model` with payload `json!({ "model_id": body.id })`
    (the registry reads `model_id`, lib.rs:242–246). Await
    `registry.v1.response.set_active_model`.
  - On reply `{ "status": "ok", "active_model": <entry> }` → return
    `Json(active_model_entry)` (200) so the caller sees the canonical
    `<capsule>:<model>` id the registry persisted.
  - On reply `{ "error": "<msg>" }` → map to `GatewayError::BadRequest("<msg>")`
    (the registry's "unknown model: ..." / ambiguity message is surfaced
    **verbatim**; the gateway does not reinterpret it). The registry owns
    resolution of `<capsule>:<model>` / bare ids and ambiguity (parent #961,
    design point B / sibling registry issue) — the gateway passes `id` through
    untouched.

### 2. Wiring

- `routes/mod.rs`: add `pub mod models;` (alongside agent/audit/... at mod.rs:15–28).
- `routes/mod.rs` `build()`, inside the `authed` block (after the agent route at
  mod.rs:130 is a natural home):
  ```rust
  .route("/api/models", get(models::list_models))
  .route("/api/models/active", get(models::get_active_model))
  .route("/api/models/active", put(models::set_active_model))
  ```
  (`get` and `put` are already imported, mod.rs:11.)
- `openapi.rs`: add the three handlers under `paths(...)` (openapi.rs:64) and
  `SetActiveModelRequest` under `components(schemas(...))` (openapi.rs:122–123);
  annotate each handler with `#[utoipa::path(...)]` mirroring the
  `#[utoipa::path]` block at `system.rs:20`.

## Acceptance criteria

- [ ] `GET /api/models` returns 200 with a JSON array of provider-entries for
      the authenticated principal (the array the registry publishes on
      `registry.v1.response.get_providers`), 401 without a valid bearer.
- [ ] `GET /api/models/active` returns 200 with the bound provider-entry **or
      JSON `null`** when nothing is bound; 401 without a valid bearer.
- [ ] `PUT /api/models/active` with `{ "id": "<capsule>:<model>" }` returns 200
      with the persisted active entry on success; returns 400 surfacing the
      registry's error message verbatim on unknown/ambiguous id; returns 400 on
      empty `id` without contacting the bus; 401 without a valid bearer.
- [ ] The gateway never resolves, enumerates, or persists model ids itself — it
      only forwards the raw `id` and relays the registry reply.
- [ ] Every model request published onto the bus is stamped with
      `caller.principal` (`with_principal`), and every reply is awaited on a
      subscription scoped to `Some(Some(caller.principal.to_string()))`.
- [ ] The three handlers and `SetActiveModelRequest` are registered in
      `openapi.rs`; `GET /api/openapi.json` lists `/api/models` and
      `/api/models/active`.
- [ ] `cargo test --workspace` green; no new clippy warnings under the core CI
      lint set (`--workspace --all-features -D warnings`).
- [ ] CHANGELOG.md `[Unreleased]` entry added.

## Tests (named regression tests)

Mirror the split already in `crates/astrid-gateway/tests/router.rs`: routes that
do **not** need a live registry are pinned in the router test harness
(`fresh_state_with_distro` + `tower::ServiceExt::oneshot`, router.rs:50+); routes
that need the bus round-trip use an in-process `EventBus` with a stub responder.

1. **`models_routes_require_bearer`** (router.rs-style, no daemon) — each of
   `GET /api/models`, `GET /api/models/active`, `PUT /api/models/active` returns
   401 with a missing/malformed bearer. Fails today (routes don't exist → 404).

2. **`set_active_model_rejects_empty_id`** — `PUT /api/models/active` with
   `{ "id": "" }` (and `{ "id": "   " }`) returns 400 **without** any bus publish.
   Assert the handler returns `BadRequest` before round-trip. Fails without the
   pre-validation guard.

3. **Bus round-trip tests** (new harness: build `GatewayState` with a real
   `Arc<EventBus>` in `event_bus`, spawn a tokio task that subscribes to the
   request topics and publishes canned `registry.v1.response.*` replies stamped
   with the requesting principal):
   - **`list_models_returns_registry_providers`** — request stamped principal `A`;
     stub replies with a 2-entry array; handler returns 200 with both entries.
   - **`get_active_model_null_is_200`** — stub replies `null`; handler returns 200
     with body `null`, not an error.
   - **`set_active_model_surfaces_registry_error_verbatim`** — stub replies
     `{ "error": "unknown model: bogus" }`; handler returns 400 with that exact
     message. Pins "gateway does not reinterpret resolution errors".

4. **`models_principal_isolation`** (threat-model regression — the load-bearing
   test). With a single shared `EventBus`: a stub responder publishes a
   `registry.v1.response.get_active_model` reply stamped for principal **B**
   only. A `get_active_model` handler invoked as principal **A** (scoped
   subscription `Some(Some("A"))`) must **time out / not observe B's reply** —
   assert it does NOT return B's entry. Then publish A's own reply and assert A
   receives it. This directly exercises the `accepts()` self-scope gate
   (`crates/astrid-events/src/route/entry.rs:168–169`). Fails if a handler
   subscribes unscoped (`None`) instead of `Some(Some(caller.principal))`.

## Threat model / security

- **Per-principal isolation is the core security property.** One principal must
  never read or set another's active model. The isolation is enforced
  cryptographically-upstream + structurally here:
  - The principal is the **verified** `CallerContext.principal` from the signed
    bearer (`auth.rs:69–85`, `verify_bearer`), never a body/query field — same
    discipline as every `/api/sys/*` handler and the
    `capsule_verb.rs:160–232` "nil source falls back to admin = RBAC bypass" note.
  - The request IPC is **stamped** with that principal (`with_principal`), which
    routes it into the caller's registry KV scope (registry state is
    principal-scoped KV, `capsules/astrid-capsule-registry/src/lib.rs:192–207`).
  - The reply is awaited on a **principal-scoped** routed subscription
    (`subscribe_topic_routed_scoped(.., Some(Some(caller.principal.to_string())))`,
    `crates/astrid-events/src/bus.rs:317`). Foreign-principal replies are dropped
    at enqueue (`route/entry.rs:168–169`, defence-in-depth at
    `route/entry.rs:187–189`), so even a buggy or malicious co-resident publisher
    cannot leak another principal's binding into this stream. **This is sound
    because** the registry's reply genuinely carries the requester's principal,
    not the registry capsule's identity — verified above via the
    `caller_context` → `publish` stamping chain (mod.rs:2106 → ipc.rs:291–295 →
    ipc.rs:272–273). Test 4 pins the end-to-end behaviour.
- **No admin/default fallback.** Unlike the kernel admin path, these handlers
  never connect under `default`/admin; an unauthenticated request is 401 at the
  `require_session` middleware (mod.rs:137–140) before any handler runs.
- **Input bound.** `PUT` body goes through `read_json_body` (64 KiB cap,
  principals.rs:567); empty `id` rejected pre-bus. The `id` is forwarded as
  an opaque string to the registry — the gateway performs no `format!`-into-topic
  interpolation with it (request/response topics are `&'static str`), so a crafted
  `id` cannot reshape the topic namespace.

## Scope boundaries (what is NOT here)

- **Per-request model override on `POST /api/agent/prompt`** — **deliberately
  excluded from v1.** A per-request `model` field would have to be *consumed* by
  the react reasoning loop, and no capsule reads it today: react's
  `handle_user_prompt` destructures `context` and branches only on
  `context.action == "cancel_turn"`
  (`capsules/astrid-capsule-react/src/lib.rs:617–632`) — it never inspects
  `context.model`. The react sibling issue plumbs only the *registry-selected*
  model into `LlmRequest.model`; it does not own a per-prompt override path.
  Shipping a `model` field on the prompt body now would forward a value that is
  silently dropped downstream — a dead feature. The override is therefore filed
  as an explicit Open Question and must NOT be added until a capsule-side issue
  owns its consumption (with a regression test that fails without the
  consumption). The reasoning seam in react is identified above for whoever picks
  it up.
- **Registry resolution, per-model provider-entries, `<capsule>:<model>`
  canonical-id stamping, ambiguity errors, single→multi-model migration /
  self-heal** — owned by the **registry sibling issue** (`capsule-registry`,
  parent #961). This gateway issue consumes that contract and surfaces its errors
  verbatim; it does not reimplement any of it.
- **`astrid models {list,current,set,unset}` CLI verb + root shorthand** — owned
  by the **CLI sibling issue** (`crates/astrid-cli`). Different surface, same
  underlying registry IPC.
- **react plumbing the registry-selected model into `LlmRequest.model`** — owned
  by the **react sibling issue** (`capsule-react`). This issue does not modify react.
- **OpenAI-shaped surface (`GET /v1/models`, `POST /v1/chat/completions`)** —
  explicitly **deferred**; see Open Questions. Not part of this issue.

## Dependencies & ordering

- **Depends on:** the registry sibling issue for the multi-model provider-entries
  and `<capsule>:<model>` canonical ids these routes return. The routes and tests
  can be built and merged against the **current** registry IPC contract (which
  already answers all three topics) — the canonical-id shape is transparent to the
  gateway (it forwards `id` and relays entries). No hard build-time dependency; a
  soft semantic dependency for end-to-end correctness of `<capsule>:<model>` ids.
- **Unblocks:** any programmatic/dashboard client needing model discovery + binding
  over HTTP (parent #961's primary use case).

## Open questions

- **Per-request model override.** Should `POST /api/agent/prompt` accept an
  optional `model` field that overrides the bound model for a single prompt? It is
  out of scope here because **no capsule consumes it** (react reads
  `context.action` but not `context.model`, lib.rs:617–632). To add it safely, a
  capsule-side child issue must first own reading `context.model` and threading it
  into `LlmRequest.model`, shipping a regression test that fails without the
  consumption; only then would the gateway forward a non-empty `model` under a
  reserved `context.model` key. File that child before re-introducing the field.
- **OpenAI-compatible surface.** Should `GET /v1/models` /
  `POST /v1/chat/completions` be added as a **separate** follow-up issue? Out of
  scope here. A thin `GET /v1/models` alias over the same `list_models` data could
  be near-free; the chat-completions endpoint is a larger surface and should be its
  own issue.

joshuajbouw:
## Linked Issue

Closes #1009

## Summary

Adds three principal-scoped HTTP routes so a script, dashboard, or automation client can discover and bind a principal's active LLM model without driving the TUI's interactive `/models` picker. The gateway is a **thin transport** over the registry capsule's existing IPC contract — it never resolves ids, enumerates models, or persists selection.

- `GET  /api/models`        — list the caller's available provider-entries.
- `GET  /api/models/active` — the caller's bound model, or JSON `null` (a valid 200, not an error).
- `PUT  /api/models/active` — bind by `{ "id": "<capsule>:<model>" }`.

## Changes

- **New `routes/models.rs`**: one `registry_round_trip` helper (publish the registry request stamped with the caller's principal, await the reply on a principal-scoped routed subscription) + the three handlers. `PUT` rejects an empty/whitespace `id` with 400 **before** touching the bus; on success it returns the persisted active entry; on a registry error it surfaces the `{ "error": "…" }` message as 400 **verbatim** (the gateway does not reinterpret resolution/ambiguity errors).
- **Wiring**: `pub mod models;` + the three routes in the `authed` block of `routes/build()`; the three handlers + `SetActiveModelRequest` registered in `openapi.rs` (and a `models` tag).

## Security notes

Per-principal isolation is the core property and is enforced structurally:

- The principal is the **verified** `CallerContext.principal` from the signed bearer — never a body/query field.
- The request IPC is **stamped** with that principal (`with_principal`), routing it into the caller's registry KV scope and making it the value the reply is matched against.
- The reply is awaited on a route scoped to `Some(Some(caller.principal))` (`subscribe_topic_routed_scoped`); a foreign-principal reply is dropped at enqueue, so a co-resident principal can neither read nor set another's active model. This is sound because the registry's reply genuinely carries the requester's principal (the host resolves the outgoing principal from the inbound caller-context). The `models_principal_isolation` test pins this end-to-end: a reply stamped for B is not observed by a handler invoked as A.
- The `id` is forwarded as an opaque string; request/response topics are `&'static str`, so a crafted `id` cannot reshape the topic namespace.

A per-request `model` override on `POST /api/agent/prompt` is **deliberately excluded** — no capsule consumes a per-prompt model today (react reads `context.action`, not `context.model`), so shipping it would forward a silently-dropped value. It must wait for a capsule-side issue that owns its consumption.

## Verification

- Code: `cargo test --workspace` passes (the only failures observed locally are two gateway TLS integration tests that bind ephemeral TCP ports, which the local sandbox blocks — they pass with the sandbox disabled and are unrelated to this change). No new clippy warnings under `--workspace --all-features -D warnings`.
- New regression tests (each fails on `main` — the routes don't exist there): `models_routes_require_bearer`, `set_active_model_rejects_empty_id`, `list_models_returns_registry_providers`, `get_active_model_null_is_200`, `set_active_model_surfaces_registry_error_verbatim`, `set_active_model_returns_persisted_entry`, and the load-bearing `models_principal_isolation`. The router OpenAPI drift check covers the new paths.

## Review fixes

Addressed a review finding that the suite-time was dominated by `models_principal_isolation`: its negative assertion (principal A observes no reply stamped for principal B) blocked for the full 10s production `REGISTRY_TIMEOUT`. The registry round-trip budget is now an optional `GatewayState::registry_timeout` override read via `unwrap_or(REGISTRY_TIMEOUT)`; production constructors leave it `None` so the 10s budget is unchanged, and the isolation test sets 300ms so its no-reply assertion resolves promptly. The proof is intact - A must still time out without seeing B's binding, then receive its own reply at 200.

## Rebase

Rebased onto current `origin/main`, which advanced to the gateway agent-loop-readiness feature (#1007). The two features are **additive and now coexist**:

- `GatewayState` gained two fields from different sides — main's `readiness_probe` and this branch's `registry_timeout`. Both are kept at every construction site (the `new()` constructor and the test struct-literals in `auth.rs`, `cors.rs`, `router.rs`, `gateway_tls.rs`, `models.rs`). The new `tests/models.rs` constructor needed `readiness_probe: None` threaded in (the rebase could not auto-fill it because the file is new on this branch).
- `routes::build()` registers **both** the `/api/models*` routes and main's `/api/sys/readiness` route; `openapi.rs` registers both surfaces' paths/schemas/tags. With both route sets present `build()` ticks one line over clippy's `too_many_lines` threshold, so the flat route-registration list is marked `#[allow(clippy::too_many_lines)]` (matching existing precedent in `astrid-cli` and `astrid-mcp`) — its length tracks API surface, not branching.
- `CHANGELOG.md` `[Unreleased]` merged cleanly with both entries (readiness and models).

No logic clash — every conflict was a both-sides-keep. `build` / `clippy --workspace --all-features -D warnings` / `test --workspace` all green afterward (modulo the two known sandbox-only ephemeral-port TLS tests), including the new models tests, the readiness-route auth test, and the router OpenAPI drift check that now lists both surfaces.

## Checklist

- [x] Linked to an issue


joshuajbouw:
Folded in the triaged review findings:

- **Fixed (Gemini):** the model registry round-trip stamped its IPC request with `Uuid::nil()` as the source id, which is the reserved `SYSTEM_SESSION_UUID`. That mis-attributes a client-originated gateway request as the system session. Now stamps a fresh per-request `Uuid::new_v4()`. Reply correlation is unaffected — replies are matched by the principal-scoped routed subscription, not by the request source id, which the existing `models_principal_isolation` round-trip test confirms (full `-p astrid-gateway` suite green).
- **Cleanup:** dropped a redundant `CallerContext` clone in the two GET handlers; the borrowed reference from `caller_from` passes straight through. `set_active_model` keeps its clone because `read_json_body` consumes the request.

**Deferred (Copilot, models.rs:112):** the concurrent same-principal reply-consumption race is real but needs a request-correlation design decision, so it is deferred pending a correlation-strategy decision and handled separately — not addressed here.

**Already fixed:** the 10s isolation-test slowness Copilot flagged was already addressed in 8ea4ddc (injectable registry timeout).

joshuajbouw:
Copilot's concurrency-race finding (`models.rs:112`) is now **resolved** via set-path correlation, no longer deferred.

**Mechanism.** Each `PUT /api/models/active` stamps its request with a fresh per-request `corr_id` (a UUID) carried in the JSON **payload** (not the IPC envelope). The registry echoes that `corr_id` verbatim in both its success and error replies. The gateway's reply-draining loop (`registry_round_trip`) now **skips** any reply whose `corr_id` is present and differs from its own, and **accepts** a reply with a matching `corr_id` *or* a reply with no `corr_id` field. The wait stays bounded by the existing `REGISTRY_TIMEOUT` via an absolute deadline, so a stream of skipped foreign replies can never extend the total wait.

This closes the window where two concurrent same-principal SETs landing on the same principal-scoped routed subscription could consume each other's reply **body**. The persisted state was already correct (the registry CAS), but the wrong reply could be surfaced to the caller; correlation fixes the wrong-response-body.

**Scope.** Set-path only. GET paths (`get_providers`, `get_active_model`) stay deliberately uncorrelated — concurrent same-principal reads are idempotent and benign — so they thread `None` and accept the first reply, unchanged. A reply with no `corr_id` is accepted on the set path too, for back-compat with a not-yet-updated registry (and any GET reply that races onto the route).

**Tests.** Unit tests cover the pure correlation predicate (matching / foreign / absent / non-string `corr_id`); an in-process `EventBus` regression test publishes a foreign-`corr_id` reply **then** the matching one and asserts the handler returns the matching binding (verified it fails without the change), plus a back-compat test that a no-`corr_id` reply is still accepted. The registry-side echo lands in the registry capsule under the same shared correlation contract.

joshuajbouw:
Folded the second-round re-review findings on this branch:

**Folded (robustness):** The PUT `/api/models/active` success path returned a 200 for any *present* `active_model` field — including an explicit JSON `null`. A registry reply of `{ "active_model": null }` would have surfaced as `200 null` instead of the 500 a malformed reply warrants (a genuine bind always carries the canonical provider-entry). The presence check now requires a non-null value, matching typed deserialization where `active_model: null` → `None`. I lifted the SET reply-shape decision into a pure `classify_set_active_reply` helper (`Bound` / `Rejected` / `Malformed`) so the explicit-null edge is unit-testable without a live bus, and added regression tests for all four shapes; the null test fails without the fix.

**Folded (cleanup):** `registry_round_trip` now takes `&PrincipalId` instead of `&CallerContext` — it only ever needed the principal id — so `set_active_model` clones just the id rather than the whole caller context (the GET callers pass `&caller.principal`). And the `registry_timeout` doc in `state.rs` now describes the 10-second default behaviourally instead of naming the private `REGISTRY_TIMEOUT` const from another module.

**Dismissed (already handled, no code change):**
- The concurrency race on the mutating PUT path (models.rs ~129/153) is already closed: each PUT carries a fresh per-request `corr_id` that the registry echoes and the round-trip filters (`reply_satisfies_corr_id`, commit 548e435), so two concurrent same-principal PUTs no longer cross reply bodies. The GET races are benign by design — same-principal reads are idempotent and intentionally left uncorrelated (`None`).
- The isolation test (`models_principal_isolation` in tests/models.rs ~443) does *not* block for 10s: it injects a 300ms registry timeout via `state_with_bus_timeout(.., Some(Duration::from_millis(300)))` (commit 8ea4ddc), and the comment there explains it explicitly. The 10s production budget is never hit in the test.

Workspace build + clippy clean; full test suite green except the two ephemeral-port TLS integration tests (`gateway_tls.rs`), which fail only under the sandbox's bind restriction and are unrelated to this change.

joshuajbouw:
Folded the final substantive review edges in `crates/astrid-gateway/src/routes/models.rs` (pushed `eb9803f..49a73ec`).

**Folded**

- **Correlation filter (correctness/security).** `reply_satisfies_corr_id` used `get("corr_id").and_then(as_str)`, which collapses a *present-but-non-string* `corr_id` to `None` and ACCEPTED it — that defeats the filter, so a malformed or forged reply could be surfaced as the matching SET response. Semantics are now explicit on all four cases: absent field -> accept (legacy/GET back-compat); present string == ours -> accept; present string != ours -> skip; present non-string (number/bool/array/object/explicit-null) -> skip as an anomalous non-match. The bounded recv deadline already caps the skip path, so an anomalous reply degrades to a timeout (500), the correct fail-safe, rather than the wrong body. The existing unit test that asserted a non-string `corr_id` is accepted is flipped to assert it is SKIPPED, and now covers numeric/bool/array/object/null shapes; the absent-accepted, matching-accepted, and foreign-skipped cases are retained.

- **Zero-budget recv (robustness).** The reply-draining loop could call `recv(Some(remaining))` with `remaining` == zero once the deadline elapsed; tokio's behaviour with a zero timeout is implementation-defined (it can poll the inner future once before the timer and surface a reply sitting in the route queue past the deadline). The loop now breaks to the timeout path (500) when no budget remains, before calling `recv`. The absolute-deadline accounting is unchanged. New integration test `zero_budget_round_trip_times_out_deterministically` asserts a zero-budget round-trip times out to 500 deterministically and never surfaces a queued reply — it fails on the un-guarded code (which racily returns 200) and passes once guarded.

- **Diagnostic (optional).** The malformed-set-reply error message now mentions that an explicit-null `active_model` is treated as absent, so the message matches the classifier.

Both fixes ship a regression test verified to fail when the fix is reverted.

**Dismissed (no code change)**

- **Re-flagged concurrency race (the get_active/set round-trip).** Already addressed. SET replies are correlated per-request via a fresh `corr_id` (548e435) so two concurrent same-principal SETs can no longer consume each other's reply body; the GET races are benign and idempotent by design (concurrent same-principal reads return the same state).

- **10s isolation test taking the full budget.** Already addressed — the isolation test injects a 300ms round-trip timeout (8ea4ddc), so the negative assertion does not block on the production budget.

- **`&caller.principal` borrow suggestions, the null-as-absent point, and the `REGISTRY_TIMEOUT` / route-isolation doc nits.** All addressed in eb9803f.

Build, workspace clippy (`--all-features -D warnings`), and the gateway/workspace test suites are green; the two ephemeral-port TLS integration tests fail only under the sandbox (socket-bind `Operation not permitted`) and pass with the sandbox disabled.

joshuajbouw:
Rebased onto current origin/main (now the agent-loop-readiness feature, #1007). All conflicts were additive both-sides-keep: GatewayState carries both readiness_probe and registry_timeout at every construction site, routes::build + openapi register both the /api/models* and /api/sys/readiness surfaces, and the CHANGELOG keeps both entries. The combined route list tipped build() one line over clippy's too_many_lines, so the flat registration list is now allow(too_many_lines) (existing precedent). Build / clippy --all-features / test all green afterward — both features coexist (new tip b79b37d).

PR #1016, opened 2026-06-23
feat(cli): provider + model onboarding selects

### LINKED ISSUE #1015: feat(cli): provider + model onboarding selects
## Problem

During `astrid init`, onboarding collects the LLM model as free-text and never frames the provider choice as a deliberate multi-select. There is no live model picker, so an operator must already know a valid model id for their provider and type it exactly. Capsule `[env]` already supports a *static* `select` (via `enum_values`), but a provider's model list is dynamic and version-dependent — a hardcoded enum goes stale.

## Proposed flow

1. **Multi-select providers** — present the distro's `group = "llm"` capsules as a multi-select ("Which LLM provider(s) do you want to set up?").
2. **Per-provider credentials** — for each selected provider, collect its capsule-specific `[env]` fields (api_key, base_url) from the capsule's own schema.
3. **Live model select** — a dynamic `options-from` `[env]` field fetches the provider's model list (e.g. `GET {base_url}/v1/models` with `Authorization: Bearer {api_key}`) and presents the ids as a numbered select. Graceful fallback to free-text on any discovery failure.

End state: each selected provider is installed and configured (creds + chosen default model), so the model registry spans them all and `astrid models` switches across the set.

## Scope

- **Manifest schema + installer code only** — no WIT contract change, so no RFC.
- Part A: extend capsule `[env]` `EnvDef` with `options-from` (http/bearer/select/after); the native installer (not the sandboxed capsule) performs the live fetch and parses the OpenAI `data[].id` shape, with graceful free-text fallback.
- Part B: distro provider multi-select + sequential per-provider `[env]` onboarding so the dynamic model select runs during install.
- Part C (optional): same select capability for distro `[variables]`.

Relates to #961

joshuajbouw:
## Linked Issue

Closes #1015

## Summary

Capstone of the model-selection work: during `astrid init`, the operator multi-selects LLM providers, then for each enters credentials and picks a default model from a **live** list (the provider's `/v1/models`). Onboarding previously collected the model as one free-text `[variables]` field — the operator had to know a valid model id and type it exactly. This is manifest-schema + installer code only; no WIT contract change, so no RFC.

### Part A — dynamic `[env]` selects (`options-from`)
- Extend `EnvDef` with optional `options-from = { http, bearer, select, after }` (kebab `options-from` accepted as an alias of `options_from`, matching the snake-case `enum_values` sibling). A field with `options-from` is a **dynamic SELECT**.
- The fetch runs in the **native installer** (it has network; the sandboxed capsule does not). At prompt time, after the `after` dependency fields are collected, the installer substitutes already-entered values into the `http`/`bearer` templates, performs `GET <http>` (sends `Authorization: Bearer <bearer>` only when the token is non-empty after trim **and** the resolved fetch host matches the configured provider host — see the adversarial note), caps the response body at **5 MB** (generous for large aggregator `/v1/models` catalogs, still bounds the OOM vector), and parses the standard OpenAI `{ "data": [ { "id": ... } ] }` shape into a deduped (stable, server order) list of non-blank ids. The `default` is the pre-selected option.
- **Graceful fallback:** any failure (unresolved template, non-http endpoint, host-mismatch 401, oversized body, network error, non-2xx, non-JSON, empty data) degrades to a free-text prompt with a hint — never hard-fails the install.
- **Field ordering** honours `after` (stable topological pass; a cycle or unknown-key dependency flushes the rest in sorted order rather than wedging).

### Part B — distro provider multi-select + per-provider onboarding
- The `group = "llm"` capsules are presented as a provider multi-select ("Which LLM provider(s) do you want to set up?"); the selection parser dedupes and preserves entry order.
- For each selected llm capsule, its own `[env]` schema prompt runs after install, so capsule-specific fields (`api_key`, `base_url`) and the dynamic `model` select resolve from the installed manifest. Shared `[variables]` still template genuinely-cross-capsule values; the prompt skips already-set keys. A missing manifest or env error for one provider is reported and skipped — it never aborts the whole install.

### Part C (deferred)
Distro `[variables]` select capability was **not** added — it is independent of the provider/model flow (Part B routes capsule-specific selects through the capsule's own `[env]`, which is where the live model picker belongs), and folding it in would only complicate this PR for no functional gain in the onboarding path. Tracked as a clean follow-up.

### Implementation note
The discovery HTTP call reuses the CLI's existing `reqwest` dependency (no new dependency). Because the install-prompt routine reads stdin synchronously and is called from both sync and async contexts, the one short discovery GET runs on a dedicated current-thread runtime on a scoped OS thread — it never touches or risks deadlocking the ambient runtime, and the prompt stays strictly sequential.

## Verification

- `cargo build --workspace` — clean.
- `cargo clippy --workspace --all-features -- -D warnings` — clean.
- `cargo test --workspace` — passes, except the two pre-existing sandbox-only ephemeral-port TLS integration tests in `gateway_tls.rs` (`127.0.0.1:0` bind blocked by the sandbox; unrelated to this change).
- New regression tests:
  - `options-from` parse + `select` default (`astrid-capsule` manifest).
  - Live-fetch response parser (pure unit): server-order, dedupe, blank/missing-id drop, unknown-hint fallback, non-JSON / missing-`data` / non-array → empty.
  - `fetch_options` rejects an unresolved template and a non-http endpoint **before** any network (fallback trigger).
  - **Credential host-binding (`should_send_bearer`):** bearer sent when the fetch host matches the configured `base_url` host (incl. case-insensitive host, explicit matching port); withheld on a foreign host, a confusable subdomain, a differing explicit port, a scheme downgrade, or an unparseable/host-less URL.
  - **Body cap (`decode_capped_body`):** within-limit decodes, over-limit errors, exactly-at-limit allowed.
  - `after`-dependency ordering: deps-before-dependent, transitive chain, unknown-dep ignored, cycle terminates.
  - Provider multi-select parsing: multi, out-of-range/garbage drop, dedupe, entry-order preservation.

## Adversarial review note

The `http` template is capsule-author-controlled (an SSRF surface on the operator's machine, bounded by and consistent with the existing install trust model — distro install already fetches author-controlled archives and runs author-controlled WASM, and the operator installs this capsule deliberately and sees the resulting prompt). The installer enforces:

- **Scheme:** non-`http(s)` endpoints are rejected before any network call.
- **Credential host-binding:** the `Authorization: Bearer <token>` header is attached **only** when the resolved fetch host equals the host of the user-configured provider `base_url` the `http` template is built from — case-insensitive host with an exact port match. On a host mismatch (or any unparseable/host-less URL) the bearer is withheld and a warning is logged; the unauthenticated request then typically 401s into the free-text fallback. This closes the credential-exfiltration path where a capsule declared `http` at an attacker host while resolving `bearer` to the user's API key. The legitimate openai / openai-compat case builds `http` from `{base_url}`, so the hosts match and the bearer is sent as before.
- **Response size:** the body is capped at 5 MB (advisory `Content-Length` pre-check plus a hard byte-length bound) — generous for large aggregator `/v1/models` catalogs (hundreds of models with metadata), still bounding the OOM vector; an over-limit body errors into the free-text fallback rather than buffering an unbounded response.

Private-IP SSRF is not blocked (a host/scheme allowlist could tighten it in a follow-up if desired), but no credential leaves the operator's machine to an unconfigured host.

## Checklist

- [x] Linked to an issue


joshuajbouw:
Folded in two review must-fixes (7e8137d):

1. **Credential host-binding (security).** `options-from.http` and `options-from.bearer` were independent templates with no host binding, so a capsule could declare `http = "https://attacker.com"` + `bearer = "{api_key}"` and the installer would ship the user's provider key to an arbitrary host. I extracted a pure `should_send_bearer(http_url, base_url)` seam and now attach the `Authorization: Bearer` header only when the resolved fetch host equals the host of the user-configured `base_url` (case-insensitive host, exact port). On a mismatch the bearer is withheld with a logged warning and the unauthenticated request falls back to free-text. The openai / openai-compat case builds `http` from `{base_url}`, so its host matches and the bearer is sent as before.

2. **Response body cap (DoS).** The models fetch read `response.text()` with no size limit. It now rejects an advertised over-limit `Content-Length` up front and hard-bounds the actual bytes at 512 KB via a pure `decode_capped_body` seam; an over-limit body errors into the existing free-text fallback.

Both are covered by no-network regression tests over the pure seams (host match/mismatch/subdomain/port/scheme-downgrade/unparseable; within/over/at-limit). Reworded the PR's adversarial-review note to state exactly what is enforced (scheme rejection, host-bound bearer, 512 KB cap) rather than the prior phrasing. `build`/`clippy --all-features`/`test` all green (modulo the two known sandbox-only TLS ephemeral-port tests).

joshuajbouw:
Raised the models-response cap from 512 KB to 5 MB. A typical `/v1/models` response is KB-scale, so the cap was never about the common case — it exists to stop a hostile or broken endpoint from OOM-ing the installer. 512 KB was tight enough to risk clipping the largest legitimate aggregator catalogs (hundreds of models with metadata) to the free-text fallback; 5 MB accommodates those while still bounding the OOM vector. The `Content-Length` pre-check and the `decode_capped_body` byte bound both track `MAX_RESPONSE_BYTES`, and the regression test now asserts the 5 MB threshold with a within-limit and an over-limit case. Build, clippy (`--all-features -D warnings`), and the workspace tests are green (the two sandbox-only ephemeral-port TLS tests aside).

joshuajbouw:
Folded the second-wave re-review findings (security + correctness), each with a regression test:

- **DoS via unbounded body (FIX A):** `fetch_options` now streams the `/v1/models` body through `bytes_stream()` and aborts the moment the running total crosses `MAX_RESPONSE_BYTES` (5 MB), so a chunked / unknown- or lying-`Content-Length` response can no longer buffer the whole body and OOM the installer. The up-front `Content-Length` reject is kept as a fast path. New `fetch_rejects_oversized_chunked_body_without_buffering_it_all` spins a raw-TCP chunked server (no `Content-Length`) and asserts both the size error and that the server never had to push a full second cap's worth of bytes before the client dropped the stream.
- **Scheme-downgrade credential leak (FIX B):** `should_send_bearer` now requires scheme to match in addition to host+port. Previously `http://provider:443` vs a configured `https://provider` folded to the same `port_or_known_default` (443) and leaked the bearer in cleartext. New assertions cover same-host same-explicit-port http-vs-https withholding (both directions).
- **Template corruption (FIX C):** `resolve_template` is now a single deterministic regex pass over `{\w+}` placeholders instead of an order-dependent HashMap-iteration string-replace, so substring keys (`{api}` vs `{api_key}`) and map order can't corrupt the URL/credential, and a substituted value containing `{...}` is never re-scanned. Tested order-independent across two insertion orders plus the no-rescan case.
- **Blank-as-default re-prompt (FIX D):** confirmed real against `write_env_files`, which deliberately writes an empty string to mean "use the capsule default." `prompt_env_fields` now skips any key already PRESENT (including explicitly-empty), prompting only for UNSET (absent) keys, so the distro-preconfigured blank-as-default fields no longer nag. Factored the decision into `should_prompt_for_key` and pinned it with a unit test plus an end-to-end "fully-preconfigured file left untouched" test.

The CHANGELOG entry for this feature was updated in place to reflect the now-accurate scheme+host+port binding and the streaming cap.

Note: the chunked-body regression test binds an ephemeral TCP port and is skipped under the build sandbox (like the existing TLS ephemeral-port tests); it passes when run with network access. Also flagging an unrelated pre-existing clippy error on this branch in `tui/state.rs:1368` (`field_reassign_with_default`, from #468, not touched by this PR) — left out of scope.

Issue #961, opened 2026-06-21
Bind a specific LLM provider/model to a principal (umbrella)


vrogojin:
# feat(registry): bind a specific LLM provider/model to a principal (umbrella)

> **Repo:** `unicity-astrid/astrid` (core). This is the **umbrella** issue and the only one that
> speaks across repos. Each per-repo work item below is tracked as its own child issue in its own
> repo; this issue links them and owns the shared design. **Core #961 closes when the two
> core-side surfaces (CLI shorthand + gateway API) land**; the capsule-side children are tracked
> and closed in their own repos.
>
> This issue **rewrites** the original #961 body. The original framing — "nothing in-repo selects
> which LLM provider/model serves a given principal" — was *almost* right but imprecise: the
> mechanism **does** exist (the `astrid-capsule-registry` capsule, per-principal) but is
> provider-granular, TUI-only, env-pinned, and decoupled from the request `react` actually sends.
> The four concrete defects are enumerated below.

## Problem / Motivation

There is no scriptable CLI or HTTP API to bind a specific LLM **provider + model** to a principal.
A programmatic integrator (e.g. an external stack provisioning an agent over the gateway) cannot,
today, say "principal *P* uses model *M* on provider *C*" and have the next prompt actually run on
*M*. The selection machinery that exists is real and per-principal, but it selects a **provider
capsule**, not a model; it is reachable only from the interactive TUI; the model string is frozen
in a capsule env var; and `react` ignores the selection when forming the request, shipping its own
env default instead.

Concretely, on a multi-provider install (say OpenAI plus a local LM Studio endpoint), selecting a
model via the only available surface changes *which topic the request is published to* but **not
the model string in the request** — so a Claude model name gets sent to an LM Studio endpoint.

## Current behaviour (verified at HEAD)

The active-model concept is owned by the `astrid-capsule-registry` capsule, whose state
(`RegistryState { providers, active_model_id }`) is persisted in its KV under one key, which the
kernel scopes **by invocation principal** — so selection is genuinely per-principal already
(`capsules/astrid-capsule-registry/src/lib.rs:52-63` for the state shape + `load_state`, `:192-208`
for the principal-scoping rationale on `handle_get_active_model`). It has four defects:

**Defect 1 — selection is provider-granular, not model-granular.** Every provider capsule
publishes exactly **one** `provider-entry` on `llm.v1.response.describe`, and that entry's `id` is
the **provider/capsule name**, not a model id:

[...]

PR #1024, opened 2026-06-23
fix(distro): enforce [distro].astrid-version against the running CLI on init

### LINKED ISSUE #1023: fix(distro): enforce [distro].astrid-version against the running CLI on init
## The gap

The distro manifest declares `[distro].astrid-version` — a CLI-version requirement (e.g. `">=0.6.0"`). The CLI parses it for well-formedness (`distro/validate.rs` runs `VersionReq::parse(...).is_err()`) but **never compares it to the running CLI version**. `init.rs::run_init` has no version check at all.

## The consequence

`astrid init` / `astrid distro apply` fetch `Distro.toml` from the repo **`main` tip** (`raw.githubusercontent.com/.../main/Distro.toml`). Because the requirement is declared and validated-as-semver but never enforced, a manifest change that bumps `astrid-version` to require a newer astrid **silently proceeds into a broken onboarding on an older CLI** — the user gets a confusing mid-install failure instead of a clear "upgrade your CLI" message. This is first hit by the #961/#1016 onboarding restructure, which moves more behaviour behind the manifest.

## The fix

Enforce `[distro].astrid-version` against the running CLI (`env!("CARGO_PKG_VERSION")` parsed as `semver::Version`) in the init/distro-apply path, **before any prompting or install**, failing with an actionable error (`This distro requires astrid {req}, but you are running {version}. Run \`astrid update\` to upgrade.`). A locally-built / dev / prerelease CLI at-or-above the floor must NOT be falsely rejected (the classic `>=0.6.0` vs `0.6.0-dev.3` semver footgun) — compare on the release triple, ignoring the prerelease tag. Keep the existing well-formed-semver validation. Out of scope: the `[distro.requires.astrid]` interface version reqs (separate, bigger concern).

Relates to #961, #1016

joshuajbouw:
## Linked Issue

Closes #1023

## Summary

The distro manifest declares a CLI-version floor via `[distro].astrid-version` (e.g. `">=0.6.0"`), but the CLI only validated it as *well-formed* semver and never compared it to the running binary. Because `astrid init` / `astrid distro apply` fetch `Distro.toml` from the repo `main` tip (`raw.githubusercontent.com/.../main/Distro.toml`), a manifest that raised its floor would pass validation and break onboarding mid-flight on an older CLI with a confusing error instead of an actionable "upgrade your CLI" message. First hit by the #961/#1016 onboarding restructure.

This adds the enforcement, gated on the running CLI version before any prompting or install.

## Changes

- New `enforce_astrid_version(&DistroManifest)` in `commands/distro/validate.rs`, wired into `run_init` immediately after `fetch_and_parse_manifest` and before lock-freshness / select / install — so both `astrid init` and `astrid distro apply` (which call `run_init`) are gated.
- The running version comes from `RUNNING_ASTRID_VERSION` = `CARGO_PKG_VERSION` (the same source `astrid version` prints), parsed as `semver::Version`.
- On a floor miss it fails fast with: `This distro requires astrid {req}, but you are running {version}. Run \`astrid update\` to upgrade.`
- **Prerelease footgun handled.** The pure, unit-testable `distro_astrid_version_satisfied(req, running)` compares on the running version's **release triple** (prerelease/build metadata stripped), so a locally-built / dev / prerelease CLI at or above the floor (e.g. `0.6.0-dev.3` vs `>=0.6.0`) is **not** falsely rejected by `semver`'s default prerelease rule — while a CLI whose release triple is genuinely below the floor still fails. Policy documented in a doc comment.
- A manifest with no `astrid-version` imposes no requirement; a malformed `astrid-version` remains a manifest error (existing well-formed-semver validation kept).
- Scope is the `[distro].astrid-version` CLI gate **only** — `[distro.requires.astrid]` interface version reqs are untouched (separate concern).
- CHANGELOG `[Unreleased]` entry. No WIT change, no version bump.

## Verification

- `cargo build --workspace` clean (pre-existing dSYM filename-collision warnings only).
- `cargo clippy --workspace --all-features -- -D warnings` clean.
- `cargo test --workspace -- --quiet` passes except the two known sandbox-only ephemeral-port TLS tests in `gateway_tls` (`Operation not permitted` on ephemeral bind — environmental, unrelated).
- New unit tests in `validate.rs` (each fails without the change): running CLI **below** floor is rejected with the actionable message; **at/above** floor accepted; a **prerelease/dev** running version at-or-above the floor is **accepted** (the footgun, with an internal guard asserting naive `VersionReq::matches` would have rejected it); a prerelease **below** the floor is still rejected (the triple-strip does not over-accept); a **malformed** requirement is still an error; a manifest with **no floor** imposes no requirement.

## Checklist

- [x] Linked to an issue


joshuajbouw:
Folded in the review fix for the version-gate firing during `astrid update`'s own post-swap sync.

**Typed error, no string matching.** `distro_astrid_version_satisfied` now returns a typed, downcastable `AstridVersionTooOld { req, running }` (`crates/astrid-cli/src/commands/distro/validate.rs`) instead of a bare `anyhow::bail!`. Its `Display` is byte-for-byte the same actionable "This distro requires astrid {req}, but you are running {running}. Run \`astrid update\` to upgrade." message, so the `astrid init` / `astrid distro apply` paths propagate it to the top-level handler and surface the identical text — UX unchanged there.

**Benign message on the post-update path.** `sync_distro_and_capsules` (`crates/astrid-cli/src/commands/self_update.rs`) re-runs init in the still-running OLD process after the new binary is already on disk; if the freshly-fetched manifest raised its floor, the gate fired and confusingly told the user to update again right after a successful update. The error now flows through a tiny pure `post_update_sync_message(&anyhow::Error)` that downcasts to `AstridVersionTooOld` and substitutes a benign "The updated distro manifest requires the new astrid; it will take effect on your next run — restart astrid (or re-run \`astrid distro apply\`)." Every other sync failure keeps the generic `Distro sync: {e}` warn, and the update itself is still reported successful (the binary swap already happened).

**Why a pure fn:** extracting the message-selection decision keeps it unit-testable without spinning up a real update.

**Tests** (both verified fail-without / pass-with):
- `astrid_version_rejects_running_below_floor` now asserts the error downcasts to `AstridVersionTooOld` with the right req/running AND that Display is still the actionable message.
- `post_update_sync_message_softens_version_gate` asserts the gate error yields the benign restart message (and not the raw "Run \`astrid update\`" text), while a non-gate error keeps the generic warn path.

Added `thiserror` (already a workspace dep) to astrid-cli for the `#[derive(Error)]`. cargo build / clippy --all-features -D warnings / workspace tests all green (except the two known sandbox-only ephemeral-port TLS integration tests, unrelated).

joshuajbouw:
Folded both second-wave re-review findings, each with a regression test:

- **Prerelease-floor over-acceptance (FIX E):** `distro_astrid_version_satisfied` now only strips the running version's prerelease to its release triple when the requirement is a *plain release floor*. When the requirement itself names a prerelease (any comparator with a non-empty `pre`), the triple-strip is dropped and the running version is compared as-is under exact semver semantics. Previously `>=0.6.0-rc.3` with running `0.6.0-rc.2` was wrongly accepted because `rc.2` got lifted to `0.6.0`. New `prerelease_requirement_compares_running_exactly_no_triple_strip` asserts `>=0.6.0-rc.3` rejects `0.6.0-rc.2` (typed `AstridVersionTooOld`) and accepts `0.6.0-rc.3` / `0.6.0-rc.4` / `0.6.0` / `0.7.0`. The plain-release dev-build footgun fix (`0.6.0-dev.3` satisfies `>=0.6.0`) is unchanged.
- **Downcast robustness (FIX F):** `post_update_sync_message` now walks `err.chain()` to find `AstridVersionTooOld` instead of inspecting only the surface error, so the post-update benign-message softening still fires if a caller wraps the gate with `.context(...)`. Extended the existing test with a doubly-`.context`-wrapped gate (outermost display is now the context string, guarded with an assert) and confirmed it still yields the "takes effect next run" message.

CHANGELOG entry for this feature updated in place to document the prerelease-requirement exception.

Issue #965, opened 2026-06-21
`astrid:http` SSRF airlock blocks local/private LLM endpoints with no documented exception


vrogojin:
_Found while integrating the **concierge** stack against AstridOS @ `cd43283` (LLM provider capsule + agent chat over the gateway). File/line refs are at that ref._

**Severity: major.**

The host `astrid:http` interface blocks private/loopback/link-local IPs (`crates/astrid-capsule/wit-staging/deps/astrid-http/http@1.0.0.wit:3-6,26`, `airlock-rejected`). A public provider (`api.moonshot.ai`, `api.openai.com`) works, but a self-hosted/LAN model (`base_url=http://localhost:1234`, `192.168.x`) is rejected. CHANGELOG #181 cites `localhost:1234`/LM Studio as a use case, implying an allow-list exists, but no per-capsule egress/airlock-exception config is documented.

**Ask:** document (or add) a per-capsule egress allow-list for local LLM endpoints.

PR #1029, opened 2026-06-24
feat(capsule): operator-consent for local-egress (transport-origin marker + elicitation + guided pre-bless)

### LINKED ISSUE #1028: feat(capsule): operator-consent for local-egress (transport-origin marker + elicitation + guided pre-bless)
## Problem

After the SSRF-airlock hardening (#1021) closed the accidental `http://127.0.0.1` IP-literal bypass, the *only* sanctioned way for a capsule to reach a loopback/private LLM endpoint (LM Studio on `127.0.0.1:1234`, Ollama, a LAN IP) is the operator allowlist `[security.capsule_local_egress]` landed in #1026 — which today requires **hand-editing `etc/config.toml`**. That is a poor onboarding experience for the local-LLM use case driving #961, and there is no runtime path to grant access on demand.

The deeper architectural gap: at the LLM-capsule egress site, the caller is identified only by `effective_principal()`; the routed `IpcMessage` carries no **transport-origin marker** distinguishing a request that arrived over the local Unix socket from one that arrived over the gateway HTTP listener. The gateway `POST /agent` drives the same react/openai-compat egress path, so a remote bearer caller reaches the same site — every candidate operator signal (`device_key_id=None`, `principal=\"default\"`, `anonymous`) collides between local and remote. Without a host-stamped, capsule-unforgeable origin, runtime elicitation cannot positively identify a *local operator*.

## Solution — two composed mechanisms, one PR

**A. Host-stamped transport-origin marker + runtime elicitation.** Add a host-only `origin: MessageOrigin { System | LocalSocket | RemoteGateway }` field to the kernel-internal `IpcMessage` (mirrors how `device_key_id` was added — host-derived internal-bus metadata, **no WIT change**; `to_wit_message()` keeps dropping it so it stays guest-invisible). Absent/unknown/legacy/CLI-proxy-wire frames resolve to `System` (fail-closed, non-local). The kernel stamps `LocalSocket` only for a kernel-bound (handshake-verified) socket connection, `RemoteGateway` at the gateway bus-publish sites, and propagates the *originating* request's origin through guest publish/fan-out (`publish_inner`) and `publish-as` — a `RemoteGateway` request flowing through react stays `RemoteGateway` and can never be elevated to `LocalSocket`. At the airlock-rejected egress arm, if `effective_origin()==LocalSocket` and the endpoint is not already pre-blessed/granted, the host elicits via the existing `request_approval` runtime-consent primitive (`ApprovalRequired` on `astrid.v1.approval`). On approve/approve_session it caches a **per-principal** session grant (`AllowanceStore`); on approve_always it also persists to that principal's `profile.network.egress` on disk. Any non-`LocalSocket` origin (System/RemoteGateway/unbound) **never elicits** and fails closed. A consent-granted endpoint re-enters the exempt path so it refuses redirects identically to a pre-blessed one.

**B. CLI guided pre-bless.** Where onboarding / `astrid capsule config` collects a provider `base_url`, detect a loopback/private/link-local endpoint and prompt the operator (unambiguously local *by construction* in the CLI process — not runtime elicitation) to write the `[security.capsule_local_egress]` exemption keyed by capsule id. Kills the hand-edit; the API-user gotcha is structurally impossible.

## Scope / invariants

- **No WIT change** — the marker is host-only (capsules never read it). If `wit/` needs editing this needs an RFC instead.
- Origin is **always** host-populated (caller-context / ingress), never a guest argument, never reset to a per-publish default.
- V1 covers IP-literal local endpoints (the airlock-rejected arm); hostname endpoints (`localhost:1234`) still require pre-bless.
- Fail-secure throughout: deny / timeout / cancel / non-local origin / unbound socket => block.
- The runtime grant is per-principal (principal A's grant never exempts B); the operator pre-bless stays capsule-keyed.

Relates to #961, #1026. References the SSRF airlock work #965.

joshuajbouw:
## Linked Issue

Closes #1028. Relates to #961, #1026; references the SSRF airlock work #965.

## Summary

A capsule can now reach a loopback/private LLM endpoint (LM Studio on `127.0.0.1:1234`, Ollama, a LAN IP) through **operator consent** instead of a hand-edited `etc/config.toml`, without weakening the SSRF airlock #1026 put in place. Two composed mechanisms ship together.

**A — host-stamped, capsule-unforgeable transport-origin marker + runtime elicitation.** The deeper gap: the gateway `POST /agent` drives the same react/openai-compat egress path a local CLI prompt does, so at the egress site a remote bearer caller is indistinguishable from a local operator (every candidate signal — `device_key_id=None`, `principal="default"`, `anonymous` — collides). This adds a host-only `origin: MessageOrigin { System | LocalSocket | RemoteGateway }` field on the kernel-internal `IpcMessage`, stamped exactly like `device_key_id` (host-derived internal-bus metadata, **no WIT change**; the guest-facing `record ipc-message` is unchanged and `to_wit_message()` keeps dropping it). The kernel stamps `LocalSocket` only for a handshake-bound socket connection (unbound stays `System`, fail-closed), `RemoteGateway` at the gateway publish sites, and propagates the *originating* request's origin through guest publish/fan-out and `publish-as` — a `RemoteGateway` request flowing through react→openai-compat **cannot be elevated** to `LocalSocket`. At the airlock-rejected egress arm for an IP-literal local endpoint, `astrid:http` elicits one-shot operator consent (`ApprovalRequired`) **only** when `effective_origin() == LocalSocket`; approve/approve_session caches a **per-principal** `AllowanceStore` grant, approve_always also persists to `profile.network.egress`. Any non-`LocalSocket` origin never elicits and fails closed. A consent-granted endpoint re-enters the exempt path (refuses redirects like a pre-bless). V1 covers IP-literal local endpoints; hostname endpoints (`localhost:1234`) still require a pre-bless.

**B — CLI guided pre-bless.** `astrid init` / `astrid capsule config` / capsule install detect a local provider `base_url` and prompt the operator (`[y/N]`) to add the `[security.capsule_local_egress]` exemption keyed by capsule id. The operator is unambiguously local *by construction* in the CLI process, so this is guided pre-bless, not runtime elicitation. Free-text/non-resolving hosts are skipped.

## Changes

- **astrid-types**: `MessageOrigin` enum (`#[serde(default)]` + `#[serde(other)]` → `System` floor) and host-only `origin` field on `IpcMessage` + `with_origin()` builder; `skip_serializing_if` keeps the legacy wire shape for system-origin frames. Test module split to `ipc_tests.rs` to stay under the 1000-line cap.
- **astrid-capsule**: `ingress_origin` on `HostState` stamped `LocalSocket` on a bound `tcp-stream.read` (cleared on non-data/unbound, in lockstep with `ingress_principal`); `publish_inner` propagates origin from caller-context (publish) / ingress (publish-as); `clear_on_return` clears it; `effective_origin()` accessor; new `consent_egress` host module (origin-gated `consent_local_egress`); `http::egress_decision_with_consent` wraps the airlock-rejected arm; `PrincipalProfileCache::persist_egress` (load-modify-save + invalidate) for approve_always.
- **astrid-gateway**: stamp `RemoteGateway` at the `agent`, `bus_admin`, `models` publish sites.
- **astrid-hooks**: `ingress_origin: None` on the hook `HostState` (never the uplink).
- **astrid-cli**: `local_egress` module (`is_local_address`, `record_local_egress` via `toml_edit`, `maybe_prompt_local_egress`); wired into `prompt_env_fields` (install + `astrid init`) and `astrid capsule config --set`.
- **CHANGELOG**: `[Unreleased] / Security` entry.

## Verification

- `cargo build --workspace` — clean.
- `cargo clippy --workspace --all-features -- -D warnings` — clean.
- `cargo test --workspace` — green except the one known sandbox-only test `model_discovery::fetch_rejects_oversized_chunked_body_without_buffering_it_all` (binds a loopback ephemeral port — `Operation not permitted` under the sandbox; passes with sandbox off; untouched by this change). astrid-capsule 481 / gateway 108 / astrid-types 38 / CLI local_egress 12 pass.
- New tests (each fails without its change): `MessageOrigin` serde floor (absent/unknown/legacy/CLI-proxy-wire → System); bound socket stamps LocalSocket, unbound stamps System; fan-out keeps RemoteGateway (no elevation); pool reset clears `ingress_origin`; LocalSocket+unlisted elicits, approve caches per-principal grant, approve_always persists to `profile.network.egress`; RemoteGateway/System/unbound never elicit (fail closed); per-principal isolation; consent grant yields the exempt host (so it refuses redirects); `is_local_address` detection + the operator-config write/idempotency.
- No `wit/` change (the marker is host-only); commit GPG-signed (`%G?`=G).


joshuajbouw:
Folded in the two review findings. The anti-forge transport-origin core (MessageOrigin marker, stamp/propagate, fail-closed) is untouched.

**FIX 1 — MEDIUM (per-capsule grant scope).** The runtime local-egress grant was keyed on principal + host:port but not the requesting capsule, so a grant for one capsule exempted every capsule that principal ran from the airlock for that endpoint.
- `AllowancePattern::NetworkHost` and `SensitiveAction::NetworkRequest` now carry a `capsule_id`; the matcher requires the action's `capsule_id` to equal the grant's. Making it a required field (not an `Option`) let the type system catch every construction site.
- The consent gate keys both the grant WRITE and the fast-path LOOKUP on principal + capsule_id + host:port.
- `approve_always` persistence moves from the flat per-principal `profile.network.egress` Vec to a capsule-keyed `profile.network.capsule_egress` map (mirrors the operator `[security.capsule_local_egress]` shape; `#[serde(default)]` so pre-existing profiles still load; the flat `egress` allowlist is left untouched). `PrincipalProfile` is kernel-internal — no WIT/RFC.
- Regression: a `react` grant reaching `127.0.0.1:1234` no longer exempts `openai-compat` reaching the same endpoint for the same principal — proven at the pattern matcher, the in-memory session grant, the persisted `capsule_egress` map, and end-to-end (react `approve_always` then openai-compat still elicits + is denied).

**FIX 2 — LOW (response-topic isolation): kept the shared topic, documented the residual.** Egress consent reuses `astrid.v1.approval[.response.*]`. I considered switching to a structurally distinct `astrid.v1.egress.response.*`, but the only surfaces that may answer a consent are the ones holding publish rights on the response topic — the operator uplink (`capsule-cli`) and the trusted MCP broker (`sage-mcp`), both of which declare `astrid.v1.approval.response.*` and live in **other repos**. A distinct namespace would silently break the consent round-trip until those out-of-repo manifests were extended in lockstep — an ACL spread across repos this change should not require. So the shared topic stays and the residual is documented in-code (`elicit_egress_consent`): it is LOW and non-escalating because those publish rights belong only to trusted operator/broker surfaces (an arbitrary untrusted capsule cannot obtain them under the subtree publish ACL), the consent is already origin-gated to `LocalSocket` upstream, and the `request_id` is an unguessable v4 UUID. Closing it fully is deferred to a paired change that also extends the `capsule-cli` operator-surface publish ACL.

Build + clippy (`--workspace --all-features -D warnings`) clean; `cargo fmt` applied. Workspace tests green except the known sandbox-only `model_discovery::fetch_rejects_oversized_chunked_body_*` ephemeral-port bind test (`bind: PermissionDenied`, unrelated to this change). No WIT/host-ABI change; no version bump.

joshuajbouw:
Fixed a correctness gap I found on this branch: `approve_always` persisted the grant to `profile.network.capsule_egress[<capsule>]` on disk, but nothing at runtime consulted it. The egress permit fast-path checked only the in-memory `AllowanceStore`, which starts empty on every daemon boot — so a remembered grant was silently ignored after a restart and the operator got re-prompted, contradicting the "remember across restarts" contract in `consent_egress.rs` and the CHANGELOG.

The fix consults the persisted per-principal, per-capsule `capsule_egress` map in the consent gate before eliciting, mirroring how the gate already consults the operator pre-bless allowlist (`[security.capsule_local_egress]`). Isolation is preserved across the persisted path: only the effective principal's profile and the current capsule's bucket are read, so a remembered grant for capsule A / principal X never permits capsule B / principal Y, and it's still reached only after the upstream `LocalSocket` origin gate and SSRF airlock (it short-circuits the elicitation, never the origin/airlock checks). A fail-closed read means an unreadable/future-version profile falls through to eliciting rather than permitting.

Added a regression test (fails without the fix): with an empty in-memory store but a persisted `react` grant for principal `alice`, `consent_local_egress` permits without eliciting, while `(alice, openai-compat)` and `(bob, react)` still elicit. Updated the CHANGELOG #1028 entry so the persistence claim is accurate. Build + clippy (`--workspace --all-features -D warnings`) clean; full workspace tests pass apart from the known sandbox-only `127.0.0.1:0` ephemeral-port bind test in `model_discovery`.

joshuajbouw:
Pushed the adversarial-review must-fixes (f025742):

**1. Closed the global SSRF env-var bypass of the consent gate.** `SSRF_BYPASS` read two env vars at runtime in production; when set, `is_safe_ip()` returned true for all IPs, so `egress_decision` returned `Ok(None)` and `consent_local_egress` was never reached — a `RemoteGateway` caller could reach loopback/private endpoints. Fix:
- `ASTRID_TEST_ALLOW_LOCAL_IP` is now gated behind `cfg(test)` (via `test_env_bypass()`, a `const false` in non-test builds). The daemon's release binary never even reads it. The existing `is_safe_ip` loopback-block tests still pass — none of them set the var, so nothing in the harness depended on it being honored.
- `ASTRID_ALLOW_LOCAL_IPS` stays recognized in production (operators/CI may rely on it) but is now deprecated, with a one-time warning honest about the blast radius (disables the airlock for ALL capsules AND exposes local/loopback endpoints to REMOTE API callers, bypassing the per-capsule consent gate) and a pointer to `[security.capsule_local_egress]`; it will be removed in a future release.

**2. Locked the anti-forge propagation invariant end-to-end.** New regression test drives a `RemoteGateway`-stamped request through the real fan-out hop chain (caller_context origin -> `publish` -> `publish_inner` -> republished bus message -> downstream caller_context) and asserts the egress site observes `RemoteGateway` and `consent_local_egress` DECLINES — the origin can never be reset to `System`/`LocalSocket` across a republish. The existing bus-level tests stopped at 'origin survives one publish'; this closes the loop to the consent decision.

**3. Documented the revocation caveat (no behaviour change).** There's no live revocation: the operator allowlist is a load-time snapshot and the persisted `approve_always` grant lives in a process-lifetime profile cache with no TTL/watcher, so a removed bless or deleted grant keeps permitting local egress until daemon restart. Stated this at the consent-gate module doc and in the CLI add-path output so an operator editing config to revoke isn't falsely reassured. (The revoke verb itself is tracked separately.)

Verified: anti-forge core and per-capsule scoping untouched. `cargo build --workspace` clean, `clippy --workspace --all-features -D warnings` clean, `--workspace` tests pass (the one failure is the known sandbox-only loopback-bind test in `model_discovery`). CHANGELOG #1028 entry updated. No WIT change, no version bump.

joshuajbouw:
Folded the four Copilot review findings into the Feature-B CLI guided pre-bless path (`crates/astrid-cli/src/commands/capsule/local_egress.rs` + CHANGELOG):

- **Non-interactive stdin hang (MEDIUM).** `maybe_prompt_local_egress` now gates on `std::io::stdin().is_terminal()`. A scripted `astrid capsule config --set ...`, a piped install, or CI no longer reads stdin (which would block or steal piped data) and writes no exemption — the operator adds the entry explicitly. The prompt logic is factored into a testable `prompt_and_record` seam.
- **Truncating config write (MEDIUM).** `record_local_egress` now writes `config.toml` atomically (same-dir temp sibling at mode 0o600 + `sync_all` + `rename`) instead of `std::fs::write`, so an interrupted write can never leave a truncated config the daemon fails to load. The `toml_edit` round-trip is preserved; the 0o600 perms now ride through the rename with no world-readable window.
- **`ip_is_local` drift (LOW).** Extracted the SSRF/local-address block-set predicate into `astrid_core::net::ip_is_blocked`, now consumed by **both** the airlock (`http::is_safe_ip`) and the CLI detector (`local_egress::ip_is_local`). Single source of truth — the prompt now fires for exactly the endpoints the runtime blocks, including `fec0::/10` site-local and the NAT64/6to4/Teredo embedded-IPv4 transition cases the doc previously claimed but missed. The env-var escape hatch stays only on the airlock side, so a test/CI var never suppresses the bless prompt.
- **Naming (accuracy).** Corrected the in-PR CHANGELOG entry: the operator config file is `config.toml`, not `astrid.toml`. (The `docs/models.md`/openai-compat-README occurrences are on main — separate follow-up.)

New unit + regression tests cover the non-TTY no-read/no-write path, the atomic write leaving no temp debris, and the `fec0::/10` + transition detection. Build, clippy (`--workspace --all-features -D warnings`), and `cargo test --workspace` are green (the one failing test is the sandbox-only ephemeral-port bind, unrelated). No WIT change. The verified anti-forge / scoping / hardening work is untouched.

joshuajbouw:
Fixed the mustFix: renamed `astrid.toml` to `config.toml` in two places in docs/models.md (lines 255 and 382). The operator config file is `config.toml` (`AstridHome::config_path` = `etc/config.toml`, `~/.astrid/config.toml`).

joshuajbouw:
Applied the four LOW-severity polish items from final review (f0a7bf9, no behavioural change to the verified anti-forge / consent / scoping / airlock logic):

1. **Teredo-client airlock test** — added a self-contained `net_tests.rs` assertion on `ip_is_blocked(2001:0:808:808:0:0:f5ff:fffe)` (public server 8.8.8.8 + obfuscated client 10.0.0.1). The client extraction arm was previously only covered indirectly via the HTTP airlock tests; now a refactor that drops it fails directly on the shared predicate.

2. **Atomic-write temp-name robustness** — `local_egress::write_atomic` used a PID-only temp name, so two same-process writers to one config.toml could share a temp path and stomp. Switched to the profile `io_impl` pattern: a `static AtomicU64` counter giving `config.toml.tmp.{pid}.{seq}`. 0o600 + sync_all + rename + cleanup-on-failure unchanged.

3. **LocalSocket trust assumption recorded** — (a) doc note at the consent origin gate stating LocalSocket trust is contingent on the uplink capability being uplink-capsule-only; (b) a `tracing::debug` at the `tcp_stream` stamp site recording the `capsule_id` that earned LocalSocket, so an unexpected stamper is observable. The capability gate remains the enforcement boundary — no change to who may stamp.

4. **Lock tradeoff annotated** — comment in `persist_egress` acknowledging the deliberate lock-held-across-IO (rare approve_always path; revisit only if it becomes a `resolve()` latency source). No locking change.

`cargo build --workspace`, `clippy --workspace --all-features -D warnings`, and `test --workspace` all green (the lone failure is the known sandbox-only ephemeral-port bind test). No wit/ change, no version bump.

PR #1045, opened 2026-06-24
feat(gateway): conversation threads over HTTP — list, fetch, manage, search, live feed

### LINKED ISSUE #972: Gateway: expose conversation threads over HTTP (list + fetch transcript)
_Found while building conversation-management features for the **concierge** integration against AstridOS @ `cd43283` (read kernel source + external `capsule-*` repos). Evidence cited at that ref._

**Severity: major.** There is no gateway route to **list a principal's conversation threads** or **fetch a thread's transcript**. The authed route table (`crates/astrid-gateway/src/routes/mod.rs:57-127`) covers auth/principals/caps/quotas/groups/invites/capsules/env/audit + `/api/agent/prompt`, but nothing for sessions. Transcripts are reachable only in-band over IPC (`session.v1.request.get_messages`), which the gateway never exposes. `capsule-session` also has no enumeration verb and no index key, and host KV `kv-list-keys` is namespace-internal — so a paired device/backend cannot discover or render existing threads via any first-class API.

**Ask:** add `GET /api/agent/sessions` (list, with metadata) and `GET /api/agent/sessions/{id}/messages` (transcript), backed by new `capsule-session` list/get verbs. This is the blocker for multi-device chat history in a client app.

### LINKED ISSUE #973: Gateway: per-principal live conversation feed (multi-device sync)
_Found while building conversation-management features for the **concierge** integration against AstridOS @ `cd43283` (read kernel source + external `capsule-*` repos). Evidence cited at that ref._

**Severity: major.** No standalone, per-principal SSE feed of conversation activity exists, so a second paired device cannot see another device's in-flight turn live. `GET /api/agent/prompt` (`crates/astrid-gateway/src/routes/agent.rs`) streams only the caller's own POSTed turn, filtered to that caller's `session_id` (`forward_event`, ~:275-297), 5-min cap. `GET /api/events` (`routes/events.rs`) is audit-only (`AUDIT_TOPIC="astrid.v1.audit.entry"`). `agent.v1.session_changed` is emitted by capsule-react but never exposed as a subscribable per-principal endpoint.

**Ask:** add e.g. `GET /api/agent/stream` — a per-principal SSE subscription fanning `agent.v1.*` for the caller's principal — so all paired devices (which act as the same principal) mirror conversations live.

joshuajbouw:
## Linked Issue

Closes #972
Closes #973

## Summary

Adds the HTTP gateway surface for conversation threads: list, fetch, manage, and search the caller's threads, plus a per-principal live SSE feed for multi-device sync. All routes are authenticated and principal-stamped — the kernel scopes `capsule-session`'s KV to the caller, so a caller only ever sees its own threads, and the path `:id` / query / body never substitute for the verified principal. Every route is a thin proxy to `capsule-session` over the in-process event bus.

## Changes

- `GET /api/agent/sessions` — paginated thread list (`limit`/`cursor`, `include_archived`, `total`).
- `GET /api/agent/sessions/:id` — one thread's metadata; `GET /:id/messages` — its transcript.
- `PATCH /api/agent/sessions/:id` — rename / archive via a metadata patch (forwards only the body keys actually sent, so absent ≠ clear; a smuggled `session_id` is dropped).
- `DELETE /api/agent/sessions/:id` — hard delete (`{ deleted }`, principal-scoped, no existence oracle).
- `GET /api/agent/sessions/search` — keyword search (`q`, `limit` default 20/cap 100, `cursor`, `include_archived`).
- `GET /api/agent/stream` — per-principal live feed (#973): `agent.v1.*` (in-flight turns) + `session.v1.event.*` (thread lifecycle), both subscribed `Some(Some(principal))` so a foreign principal's events are dropped at the bus before entering the stream. Not session-scoped (fans all of the caller's threads).
- Mixed-fleet 501 gate on the 1.1 verbs (list/get_meta/update/delete/search return a clean `501` when no session capsule implements them, instead of timing out); the transcript route stays ungated (its `get_messages` verb exists in 1.0).
- New `NotImplemented` (501) error variant; an in-process `CapsuleTopicProbe` (cap-free, every-caller) backs the gate. Count-only metrics; all routes + schemas registered in OpenAPI.

## Test Plan

- `cargo test -p astrid-gateway` — 141 tests, including live-`EventBus` request/reply round-trips and the #973 cross-principal isolation test (a foreign-principal event is never delivered to a principal-scoped feed).
- `cargo fmt --all -- --check`, `cargo check --workspace --all-features`, `cargo clippy --workspace --all-features -- -D warnings` — all clean.
- Not yet runtime-verified on a live daemon (HTTP routes + live feed exercised by unit/integration tests, not an end-to-end run).

The gateway proxies by string topic + serde, so it builds independently of the `astrid-bus:session` 1.1.0 WIT pin; it consumes the new `session.v1.*` verbs shipped in the capsule-session + wit changes.


joshuajbouw:
Thanks — addressed in e762468:

1. **list `include_archived` + `total`** — real gap, fixed. `SessionListQuery` now accepts `include_archived` and forwards it; `SessionListResponse` surfaces `total`. The capsule already supported both verbs end-to-end; the gateway just wasn't exposing them. Tests + OpenAPI updated.
2. **CHANGELOG `GetCapsuleMetadata`** — correct, it was a stale entry. The gate uses the in-process, capability-free `CapsuleTopicProbe` (deliberately *not* the cap-gated `GetCapsuleMetadata`, which would 403 ordinary callers and leak the capsule inventory). CHANGELOG corrected.
3. **`capsule_timeout` hardcoded constant** — fixed; it now formats the error with the actual `timeout` argument.
4. **probe double-allocation** — fixed. `capsule_topic_probe` now short-circuits on the first subscribing capsule via a new `readiness::manifest_subscribes_topic` predicate, dropping both `Vec` allocations.
5. **`Duration::from_hours`** — kept as-is. It is stable on the pinned toolchain (Rust 1.95), and `from_secs(60*60)` actually *fails* CI here: clippy's `duration_suboptimal_units` lint (a hard error under `-D warnings`) requires `from_hours(1)`.

joshuajbouw:
Round 2 — addressed in 8058b57:

- **`ensure_list_supported` 501 message / naming** (sessions.rs) — fixed. The gate covers `get_meta`/`update`/`delete`/`search` as well as `list`, so it's renamed `ensure_session_mgmt_supported` and the 501 message no longer says "listing" (which read wrong on `PATCH`/`DELETE`/`search`) — it now names the 1.1 conversation-management verb set. It still probes the `list` verb specifically as a proxy, since the whole set ships in one capsule.
- **`GatewayState.topic_probe` doc** (state.rs:258) — fixed; it now lists all the routes that gate on it, not just `/sessions`.
- **`stream.rs` `to_string(&ready).unwrap_or_default()`** — kept as-is. It mirrors the established SSE ready-event pattern in `agent.rs` (lines 276 and 449), and `FeedReady` is a single-`String` struct whose serialization is infallible, so the empty-default branch is unreachable. Switching just this one site to `expect` would diverge from the sibling routes for no real gain; consistency wins here.

PR #1064, opened 2026-06-24
feat(capsule): capsule-set epoch + astrid:sys@1.1.0 host fn

### LINKED ISSUE #982: Runtime capsule install never invalidates non-default principals' tool-schema cache
_Found while integrating first-party tool capsules for the concierge stack against AstridOS @ `cd43283`. File refs at that ref + the published `capsule-prompt-builder`._

**Severity: major.** A capsule installed at runtime (`astrid capsule install`, or `POST /api/capsules`) has its tools offered to the **default** principal but **never** to already-active **non-default** (gateway-minted) principals — permanently, even across daemon restarts.

### Mechanism
- The LLM tool list is a describe fan-out done per-prompt by `capsule-prompt-builder` (`collect_tool_schemas()`), cached in the **per-principal** KV key `__tool_schema_cache`; on a cache hit it returns the cached list and does not re-run the fan-out.
- The cache is only invalidated by the prompt-builder's `on_capsules_loaded` interceptor (→ `kv::delete(__tool_schema_cache)`), which fires on the kernel's `astrid.v1.capsules_loaded` broadcast.
- The kernel publishes `astrid.v1.capsules_loaded` with **no principal** (`astrid-kernel/src/lib.rs` ~`load_all_capsules()`; `IpcMessage::new(...)` sets `principal: None`, never `.with_principal(...)`). The dispatcher routes an event to a single per-(capsule, principal) queue keyed by the message principal (`astrid-capsule/src/dispatcher.rs` ~`m.principal.clone()`), so the interceptor runs **only in the default/None principal context** and clears only the default principal's cache.
- KV is namespaced `{principal}:capsule:{capsule_id}` (`astrid-capsule/src/engine/wasm/host_state.rs`), so each gateway-minted principal keeps its own `__tool_schema_cache`, populated on its first prompt and never invalidated by a runtime install. The cache is persistent (survives restart).

### Impact
Backend-mediated principals (the whole A1/concierge model — every user is a gateway-minted non-default principal) never pick up runtime-installed tools. Build-time-baked capsules work only because they exist at first boot, before any principal has cached.

### Repro
1. Boot daemon; have a non-default principal P send one prompt (caches its tool set).
2. `astrid capsule install <a tool capsule>`; `POST /api/sys/capsules/reload`; restart the daemon.
3. P still never sees the new tool; the `default` principal does.

### Ask
Fan `capsules_loaded` out to all known principals (or make `invalidate_tool_cache` principal-agnostic / clear all `*:capsule:capsule-prompt-builder:__tool_schema_cache`), and/or expose a gateway endpoint to invalidate a principal's tool-schema cache. Today the only ways to refresh an existing principal are recreating it or deleting its cache KV entry out-of-band.

### LINKED ISSUE #1063: Expose a capsule-set epoch (astrid:sys@1.1.0) so capsules can self-heal load-derived caches
## Problem

Capsules that cache data derived from the *set* of loaded capsules — most notably `capsule-prompt-builder`, which caches the LLM tool list from a per-prompt `tool.v1.request.describe` fan-out in per-principal KV — have no cheap way to learn that the loaded set changed (a runtime `astrid capsule install`/reload/unload, or a hot-swap upgrade). Today the only signal is the principal-less `astrid.v1.capsules_loaded` broadcast, which the dispatcher routes to a single `(capsule, None)` queue, so an interceptor keyed on it only ever fires in the default principal's context. Backend-mediated (non-default) principals never get invalidated. This is the root cause of #982.

## Proposal

Add a kernel-side, read-only introspection primitive a sandboxed capsule can call in its **own** principal context, at request time, to detect that the loaded set changed — then re-derive only when it did. No per-principal fan-out, no kernel knowledge of any consumer's cache.

- `CapsuleRegistry::set_epoch() -> CapsuleSetEpoch`: a stable hash over the sorted `(capsule_id, package_version)` set. Moves on install/removal/version-bump, load-order independent, carries no process state (restart-stable for an unchanged set).
- New host call `capsule-set-epoch: func() -> u64` on a new **`astrid:sys@1.1.0`** (additive superset of `@1.0.0`; the host serves both, the SDK binds latest). Infallible — `0` if the registry is momentarily unavailable, biasing a consumer toward a refresh.
- SDK wrapper `astrid_sdk::runtime::capsule_set_epoch() -> CapsuleSetEpoch`.

A consumer stores the epoch beside its cache and re-validates on its next turn. Kernel stays dumb (reports a fact about its own loaded set).

Part of the #982 fix (the prompt-builder cache rewrite consumes this).

joshuajbouw:
## Linked Issue

Closes #1063

## Summary

Exposes a **capsule-set epoch** — a stable fingerprint of the loaded capsule set — so a sandboxed capsule can detect that capsules were installed / removed / upgraded and self-heal a load-derived cache, without the kernel fanning a per-principal invalidation out to every principal. This is the kernel primitive behind the #982 fix (the `capsule-prompt-builder` cache rewrite consumes it; that lands separately and closes #982).

## Changes

- `CapsuleRegistry::set_epoch() -> CapsuleSetEpoch`: a hash over the sorted `(capsule_id, package_version)` set, wrapped in a `CapsuleSetEpoch` newtype (serde-transparent, `Copy`, `Display`). Moves on install/removal/version-bump, load-order independent, carries no process state → restart-stable for an unchanged set.
- New **`astrid:sys@1.1.0`** host interface: a full additive superset of the frozen `@1.0.0`, registered in the wasmtime linker alongside it and backed by one impl — the `@1.1.0` carry-over methods delegate to the `@1.0.0` path through exhaustive (no catch-all) type conversions, so existing capsules are untouched and a new capsule binds only the latest yet reaches the whole `sys` surface (the `astrid:http@1.1.0` pattern). Adds `capsule-set-epoch: func() -> u64` — read-only, ungated like `enumerate-capabilities`, infallible (`0` when the registry is momentarily unavailable, biasing a consumer toward a safe refresh).
- Bumps the `wit` submodule to the `sys@1.1.0` superset and stages it.

## Verification

- `cargo test -p astrid-capsule` — 5 registry epoch tests (install→moves, restart-stable, load-order independent, version-bump, empty-deterministic), 2 host-fn tests (delegates to registry; `0` with no registry), 1 `@1.1.0`↔`@1.0.0` delegation-agreement test. All green.
- `cargo clippy -p astrid-capsule --all-features` clean.
- No `@1.0.0` shape change — frozen and untouched.

## Checklist

- [x] Linked to an issue

> Depends on `unicity-astrid/wit` PR (the `sys@1.1.0` superset); the submodule pin points at that branch and re-pins to its merge commit.


joshuajbouw:
Closing in favour of #1069 (per-principal capsule loading). The capsule-set-epoch primitive assumed a single global loaded capsule set; but principals are on per-principal distributions and versions (loaded per-principal off their own meta.json by content hash). With per-principal loaded sets the tool surface is per-principal by construction — no global epoch, no cross-principal cache — so this primitive is the wrong layer. Superseded; tracked under #1069.

PR #1072, opened 2026-06-25
Per-principal content-addressed capsule loading — full principal isolation of the capsule/tool surface

### LINKED ISSUE #1069: Per-principal capsule loading (content-addressed, off per-principal meta.json) — full principal isolation of the capsule/tool surface
## Problem

Capsule loading is **global**, but principal isolation requires it to be **per-principal**. Three symptoms, one root cause:

1. **Cross-principal disclosure (security).** The per-principal access gate (#993) covers only the *user-invocable* surface (`tool.v1.execute.*`, `cli.v1.command.execute`, `dispatcher.rs:216`). **Visibility is ungated:** a non-default principal's `tool.v1.request.describe` fan-out reaches every loaded tool capsule (including the default/admin principal's) and collects their tool schemas; and `materialize_principal_furniture` (#940) mirrors the default principal's *entire* `.local/capsules/` + `wit/` into every principal's home unfiltered. So a non-default principal can **see** the default/admin principal's capsules, manifests, interfaces, and tool definitions — it just can't execute them.
2. **No per-principal versions/distros.** The registry is keyed by capsule **name** (`registry.rs`, one `HashMap<CapsuleId, Arc<dyn Capsule>>`), `load_all_capsules` discovers only `PrincipalId::default()`'s `capsules_dir`, and **every** install writes to the default principal (`lifecycle.rs:109`). So exactly one version of each capsule is loaded, for everyone — two principals cannot be on different versions or different sets.
3. **#982** (tool-schema cache never invalidates non-default principals) is a downstream symptom: the prompt-builder maintains a per-principal cache of a *global* describe fan-out, and there's no per-principal invalidation signal that reaches every principal.

## Model (the fix)

The storage layer is **already built for this** — it just isn't wired into the loader:
- `CapsuleMeta` / `meta.json` (`astrid-capsule-install/src/meta.rs:22`) already records, per capsule: **name + version + `wasm_hash` (BLAKE3) + WIT blob hashes + config + details**.
- WASM blobs are already **content-addressed** at `bin/<hash>.wasm` (shared, deduplicated); `meta.json` references them by hash. WIT likewise.

Wire the **loader** to honour it:
- **Per-principal load.** For each principal, read *their* `meta.json` set and load the exact bytes each entry pins, resolved by hash from the shared content-addressed store.
- **Content-addressed instance store + per-principal view.** Key loaded instances by **content hash** (so a hash referenced by N principals loads **once** — dedup), with a per-principal `name → hash` view. Different principals on different hashes get different running versions; identical hashes share.
- **Per-principal install** with an optional **granular copy-from-default seed** (capsules + config + other `meta.json` details) for convenience; the principal's own `meta.json` is then the tracked, divergeable record.
- **Visibility falls out.** Describe fan-out and home furniture become per-principal *by construction* — a principal can only see/describe/run what's in its own `meta.json`. No separate describe-gating or furniture-filtering hacks needed.

## Outcome

- Full principal isolation: a principal cannot see or reach another principal's (incl. default/admin) capsules, manifests, interfaces, or tools.
- Per-principal versions/distros via content-addressed hashes, with binary dedup.
- **#982 dissolves**: per-principal loaded sets ⇒ per-principal tool surface ⇒ no cross-principal cache to invalidate, no global "epoch" needed.

## Supersedes / closes

- **Supersedes #1063** (the `capsule-set-epoch` host primitive) — that was built on the now-invalidated "one global set" assumption; closing it, and the `astrid:sys@1.1.0` epoch PRs (wit, core, sdk) are being closed in favour of this.
- **Fixes #982** on completion.

## Notes

- This touches the loader, registry, instance pool, install path, and the visibility surface — RFC-relevant against host-abi (rfcs#22) and the cargo-like-manifest / BLAKE3-lock model (rfcs#26). Design captured here; RFC update to follow as the contract surface settles.
- Aim: as few PRs as possible — the change is core-local (loader + registry + install + dispatch view), so ideally one cohesive core PR.

joshuajbouw:
## Linked Issue

Closes #1069

## Summary

Makes capsule loading **per-principal and content-addressed**, closing a cross-principal disclosure: previously every principal could **see and describe** the default/admin principal's capsules and tool definitions (it just couldn't execute them — #993 gated execution but not visibility). The kernel loaded one **global** registry keyed by capsule name, discovered only from the default principal's dir, with that whole set mirrored read-only into every principal's home.

The registry is now two orthogonal layers — a **content-addressed instance store** keyed by WASM BLAKE3 `wasm_hash` (each distinct binary compiled once, shared across principals, ref-counted) and a **per-principal view** (`PrincipalId → CapsuleId → wasm_hash`) built from that principal's own `meta.json`. A principal can only see/describe/run capsules in its own view; different principals can run **different versions** (distinct hashes, both loaded, isolated). Visibility isolation is emergent from the view-scoped dispatcher — no separate gating hack.

This is core-local: **no WIT / host-ABI / SDK / RFC change.** It supersedes the abandoned `capsule-set-epoch` approach (#1063, closed) — a global epoch can't represent per-principal distributions/versions.

## Changes

- **Registry** (`registry.rs`): content-addressed `instances: HashMap<WasmHash, InstanceEntry{capsule, refcount}>` (dedup; dropped only on last release) + per-principal `views`; `WasmHash` newtype (domain-separated synthetic key for non-WASM capsules); `register`/`unregister`/`get(principal,id)`/`list(principal)`/`all_instances`/`find_instance_by_uuid`/`register_existing` (dedup).
- **Dispatch** (`dispatcher.rs` + new `dispatcher_queues.rs` sibling — split to stay under the 1000-line cap): a single `candidate_capsules` choke point — the per-principal **view is a fail-closed floor** for the user-invocable + `tool.v1.request.describe` surface (absent/empty/`anonymous`/unknown caller → empty set, **never a fallback to default's view**); the existing grant resolver applies on top (view FIRST, grant SECOND); the orchestration mesh + tool-result delivery stay **global**.
- **Load** (`lib.rs`): per-principal discovery off each principal's `meta.json`, eager for default at boot + lazy `ensure_principal_loaded` on first connect; `load_capsule`/`reload`/`unload`/`restart` take a principal; dedup so a shared hash compiles once.
- **Install** (`astrid-capsule-install`, `kernel_router/install.rs`): `InstallOptions.target_principal` (default unchanged for the CLI); install writes the blob to the shared content store + the **caller's** `meta.json` and loads into the caller's view.
- **Provisioning + furniture** (`furniture.rs`, `agent_create_helpers.rs`): furniture flips from a recurring destructive full-mirror to a **one-time provisioning seed** (granular copy from default, then divergence; never clobbers a diverged set).
- **Router** (`kernel_router/mod.rs`): `ListCapsules`/`GetCommands`/`GetCapsuleMetadata` scope to the caller's view (these leaked names/commands).

## Verification

- `cargo build --workspace`, `cargo clippy --workspace --all-features -- -D warnings`, `cargo fmt --all --check` — clean. 771 tests pass (sandbox-only loopback/Unix-socket bind failures excepted, green with sandbox off).
- New isolation tests: views isolated (A never reaches B's or default's capsules), unknown principal → empty view with no default fallback, per-principal versions load/dispatch isolated, dedup ref-count shares one instance and drops on last release, provisioning seed-then-divergence never clobbered, mesh/lifecycle topics still reach all instances regardless of caller.
- Adversarial review confirmed the two invariants in `candidate_capsules`: **fail-closed with no silent default fallback**, and **mesh-stays-global** (only the principal-stamped user-invocable + describe surface scopes by view).
- Zero migration: existing default-principal installs load as default's view unchanged.

## Follow-ups (not in this PR)

- **#982** (prompt-builder tool-schema cache staleness) is now trivial and much narrower: with per-principal loading the describe surface is already per-principal, so the remaining work is invalidating a principal's cache on **its own** capsule change (a per-principal event the kernel can now stamp) — a small prompt-builder change, separate PR.
- Optional hardening: `publish_capsules_loaded` still embeds the global tool hint, consumed only by trusted operator-local tooling (`astrid mcp serve` / TUI), not the HTTP-gateway path backend principals use — could be stripped to a bare pulse once the broker's view-scoped re-derivation is confirmed sole-source.

## Checklist

- [x] Linked to an issue
- [x] CHANGELOG.md updated (entry under `[Unreleased]` › Security)


joshuajbouw:
Closing as #1083 supercedes it given the hardening of API / CLI.

PR #1079, opened 2026-06-28
fix(http): bound the egress security gate with a timeout (fixes #1078)

### LINKED ISSUE #1078: P0: retire legacy astrid:http with a fail-closed tombstone; migrate fetch_url to astrid:net
## Summary

The agent-facing **`fetch_url`** tool **hangs indefinitely** on the deployed astrid (`ASTRID_REF=c511f5f`). Any turn in which the agent calls `fetch_url` never returns — it only ends when the *downstream consumer's* tool-stall watchdog gives up (in Concierge's case, `CHAT_TOOL_STALL_MS=120s`, which then emits `tool_stalled`). This silently breaks every flow that fetches a web page (e.g. reading a web-search result page).

This is a **critical regression**: it doesn't surface in unit tests (same failure mode as the concierge#83 capsule-access-gate incident — green tests, broken live), and it turns a common agent capability into a guaranteed turn-killer.

## Impact

- Any agent reasoning that opens/reads a URL via `fetch_url` hangs the whole turn (~120s wasted, then a forced stall/abort, no answer).
- Intermittent at the *suite* level only because it depends on whether the model **chooses** to call `fetch_url`; **when `fetch_url` is invoked, it hangs every time** observed.

## Evidence / reproduction

Surfaced by the Concierge end-to-end harness (concierge#84) driving the **live** stack as virtual devices. Captured SSE milestones:

```
search attempt 1 weak (stalled=fetch_url timedOut=false fired=true grounded=false) — retrying
GAP: search turn persistently stalled/timed out (stalled=fetch_url timedOut=false)
```

`fetch_url` fired, never produced a `tool_result`, and the 120s watchdog fired `tool_stalled` — on **both** attempts.

## Diagnosis — it is NOT a network/egress problem

The same URL is reachable instantly from every layer **except** `fetch_url`:

| From | Result |
|---|---|
| Host | `https://nodejs.org/en` → **200 in 0.12s** |
| **Inside the `concierge-astrid-1` container** | `https://nodejs.org/en` → **200 in 0.08s** |
| `browser_navigate` (playwright-mcp) | reads `https://example.com` H1 fine |
| **`fetch_url` agent tool** | **hangs → tool_stalled @120s** |

So the astrid container has working outbound egress and a sibling page-fetch capability (`browser_navigate`) works. The hang is in the **`fetch_url` tool implementation itself** (the HTTP-tools capsule), not connectivity, DNS, the airlock blocking public hosts, or the target site.

## Hypotheses for the maintainers

- `fetch_url` awaits something that never resolves (missing timeout / response-body read that never completes / stream not closed), independent of the socket connecting.
- A regression around the rmcp 1.8 API change absorbed in `c511f5f`, or in the HTTP-tools capsule (cf. closed #418 "HTTP Tools Capsule").
- No internal timeout on the tool, so a slow/odd response (or a post-connect read) blocks forever rather than erroring.

## Environment

- `ASTRID_REF`: `c511f5f`
- Stack: Concierge ⇄ AstridOS (gateway at `astrid:2787`), backend tool-stall watchdog `CHAT_TOOL_STALL_MS=120000`.
- Egress confirmed working from the astrid container (see table).

## Asks

1. Add a hard internal timeout to `fetch_url` so it errors instead of hanging (immediate mitigation — a clean error is vastly better than a 120s turn-killer).
2. Root-cause the indefinite await and fix.
3. Consider a smoke test that exercises `fetch_url` against a known-fast URL so this can't regress silently again.

Refs: concierge#84 (the e2e harness that caught this), concierge#83 (the analogous silent live-only regression class), astrid#418 (HTTP Tools Capsule).

vrogojin:
## Linked Issue

Closes #1078

## Summary

The `astrid:http` outbound request could hang indefinitely: in `send_one_hop` the async per-capsule egress-authorization gate (`check_http_request`) was awaited with **no deadline**, while the request header/total deadlines wrapped only the subsequent `send()`. A gate that never replies, or a saturated host IO semaphore before the gate runs, blocked the whole request until the consumer's tool-stall watchdog fired.

This is a hang, not a network reachability failure: the reported `fetch_url` case hung ~120s on `https://example.com`, while the same URL resolved quickly from the host and via browser navigation.

## Changes

- Wrap the egress gate, including shared I/O permit acquisition, in `tokio::time::timeout` bounded by the request's own `header_deadline`.
- Fail closed on gate elapse with `ErrorCode::Timeout`; preserve `ErrorCode::CapabilityDenied` when the gate denies the request.
- Emit `astrid.audit.http` warnings for gate denials and gate timeouts with method/scheme/host/port fields so missing `net` declarations are visible in logs.
- Avoid logging full URL query strings in the production manifest-gate HTTP denial reason.
- Add regression coverage for a never-returning gate, exhausted I/O semaphore, and permission denial path.
- Add `vrogojin` to `.github/contributors.yml` as `astrinaut` so the PR author is recorded in the contributor manifest.
- Clean noisy numbered regression-section comments from the touched HTTP regression file.
- Update the `[Unreleased] -> Fixed` changelog entry.

## Verification

- `cargo test -p astrid-capsule http_security_gate -- --quiet`
- `cargo test -p astrid-capsule engine::wasm::host::http::regression -- --quiet`
- `cargo check -p astrid-capsule --tests`
- `cargo clippy -p astrid-capsule --tests -- -D warnings`
- `cargo fmt -p astrid-capsule --check`
- `git diff --check`

## Reproduction

Applied only the new stalled-gate regression test to `origin/main` in a temporary worktree and ran:

`env CARGO_TARGET_DIR=/Users/joshuaj.bouw/dev/astrid.worktrees/agent6/core-pr1079/target cargo test -p astrid-capsule http_security_gate_stall_is_bounded_by_request_deadline -- --quiet`

Result on old code: test failed after 2.01s with `Elapsed(())`, proving the old backend waited past the request deadline before the PR fix.

## Attribution

Follow-up commit includes `Co-authored-by: vrogojin <23746834+vrogojin@users.noreply.github.com>`.

## Checklist

- [x] Linked to an issue
- [x] Regression test added


joshuajbouw:
@copilot add this user as a contributor to pass the CI please, to this PR. There is a yml file.

joshuajbouw:
I pushed a follow-up that adds the regression coverage and records @vrogojin in `.github/contributors.yml` as `astrinaut`.

For the contributor-gate failure: CI classified the PR author as `new` because they were not in the contributor manifest. There are two ways to clear that gate:

- for a one-off new contributor PR, a maintainer can add the `newcomer-approved` label;
- for an ongoing contributor, add the author to `.github/contributors.yml` at the least-privileged tier that matches the touched paths.

For this PR I used the second path. `astrid-capsule` is not in the workflow’s current core/security path lists, so `astrinaut` is enough here without over-promoting. The PR branch now includes that manifest update, plus the HTTP gate regression tests and audit-log cleanup.

vrogojin:
Closing for now — not merging upstream at this time.

Context: filed while diagnosing the concierge fetch_url hang (#1078). The **actual** cause was concierge-side — `astrid-capsule-http` was simply not in concierge's per-principal capsule-grant list, so the #993 capability gate denied the capsule's `tool_describe` and `fetch_url` was never a routable tool. Fixed in concierge (grant added; verified live: fetch_url returns a real HTTP 200 in ~0s).

This PR addresses a **separate, real, latent** issue: the egress security gate in `send_one_hop` (`check_http_request`) is awaited with no timeout, so a genuinely hung gate could still block a request indefinitely. Defensive hardening, not required for #1078. We've deployed it locally as a kernel patch on our pinned `ASTRID_REF`, so we don't need it merged on our timeline.

Reopen / cherry-pick if maintainers want it upstream — small diff (wrap the gate in `tokio::time::timeout` bounded by the request's header deadline), build-verified with `cargo check -p astrid-capsule` on rust 1.95.

PR #1083, opened 2026-06-29
fix(core): isolate capsule loading by principal

### LINKED ISSUE #1069: Per-principal capsule loading (content-addressed, off per-principal meta.json) — full principal isolation of the capsule/tool surface

joshuajbouw:
## Linked Issue

Closes #1069

## Supersedes

Supersedes #1072. This PR re-authors the same core goal on top of current `main`, but corrects the runtime-sharing model after review: same-hash capsules share the content-addressed install/artifact identity, while each principal gets its own loaded runtime instance so resolved env, KV, provenance, and lifecycle remain tenant-isolated.

## Summary

Capsule loading is now principal-view aware without allowing default/global capsule visibility to bleed into other principals:

- identical capsule hashes reuse the same content-addressed artifact on disk
- loaded capsule runtimes are keyed by principal + hash, not hash alone
- each principal sees only the capsules in its own view
- unload/restart tears down the caller principal's runtime without suppressing work because another principal has the same capsule ID
- health restart probes and backoff are scoped to the failing principal runtime instance
- user-invocable dispatch, `tool.v1.request.describe`, and `llm.v1.request.describe` use the caller's principal view before grant checks
- unprincipaled LLM describe fan-out fails closed instead of falling back to every loaded provider runtime
- principal-stamped orchestration and result topics dispatch through the caller's runtime view instead of default/global views
- session management routes wait for the caller's exact session capsule runtime to finish async warm-up before publishing request/reply IPC
- inventory, metadata, commands, install, reload, and unload all carry the caller principal through the kernel path
- inventory/metadata/command/readiness reads answer from visible manifest inventory plus loaded fallback entries, so reads do not synchronously boot capsule runtimes
- full capsule reload requests are accepted synchronously and schedule the heavy reload work in the background
- `agent.modify --add-capsule` self-heals by materializing an existing default install into the target principal's install set, then warms that principal asynchronously
- gateway session and model routes derive trusted content-addressed capsule source UUIDs from installed `meta.json` hashes, so per-principal runtime replies pass while forged same-topic replies are still ignored
- gateway capsule source IDs are now named by scheme version: `v0` is the package-only compatibility fallback, while `v1` is principal + capsule + content hash; `meta.json` is read leniently for `wasm_hash`; only metadata without a usable hash self-derives the runtime-compatible synthetic content hash from `Capsule.toml` before falling back to `v0`
- per-invocation env overlays refresh for owner and non-owner principals, so gateway onboarding/config writes take effect without requiring capsule reloads
- runtime E2E pins capsule-registry to unicity-astrid/capsule-registry#24 (`9058229f0c1f1bc634c52c62dc7a258050413072`) so concurrent provider discovery dedupes duplicate model entries
- empty-catalog LLM runtime coverage retries `/api/models` until the newly invited principal's provider discovery returns the fallback model
- runtime HTTP readiness waits now allow slower per-principal capsule loading during restart/secondary-home boot while preserving bounded failure diagnostics
- live host-call cancellation smokes now unload and restore the same principal runtime view they invoke, matching per-principal capsule inventory semantics
- CLI `capsule run` now observes `astrid.v1.capsules_loaded` while waiting for a command result and fails with a bounded cancellation error if the provider unloads first
- the capsule access gate now treats the actual `cli.v1.command.run.<provider>` topic as the user-invocable CLI command surface
- single-capsule reload/unload IPC is authorized as `self:capsule:*`, matching the caller-scoped runtime view it actually mutates
- CLI `capsule remove` validation, disk deletion, and purge now target the process principal, so `astrid --principal X capsule remove ...` removes X's install set before live unload assertions
- CLI `capsule install` and `capsule update` now also target the process principal, so operator-owned capsule artifact lifecycle checks exercise the same principal-scoped install set end to end
- async capsule loading now skips registration if the source manifest disappears during load, preventing a stale full reload from resurrecting a just-removed principal view
- gateway trusted capsule source IDs now accept `v1` source IDs only for the caller principal while retaining `v0` compatibility fallback for metadata without a usable hash
- `astrid.v1.capsules_loaded` is now emitted as principal-stamped per-principal inventory, and CLI command waiters ignore other-principal frames before deciding a provider was unloaded
- full capsule reload requests now coalesce while a reload is already in flight, with a panic-safe guard reset
- `agent.create --clone-from` now materializes the cloned profile's capsule grants into the new principal's install set, strips copied capsule env files, and warms the created principal so backend-created agents can use their cloned tool set without a restart
- cloned-principal seeding uses explicit per-capsule install materialization, while the older public introspection mirror API remains present for compatibility and is not used by production principal creation
- principal capsule loads are serialized so concurrent admin-driven warms cannot stampede WASM component construction and starve unrelated gateway/auth requests
- daemon readiness now waits on the boot-critical default/system capsule view, then warms persisted non-default profile views asynchronously through the serialized load queue
- profile warm-up loads uplink capsules for all persisted principals before full views, so registry/model routes recover quickly after restart while heavier capsules continue warming
- model registry routes actively warm the caller-scoped registry subscription before publishing request IPC, preventing post-restart model requests from being lost while profile views are still warming
- session management routes use the same warm-capable caller-scoped topic probe before returning 501, so post-restart session list waits for the caller session capsule view to load

This keeps disk/content reuse through hash-addressed artifacts while preserving the accounting/security model: the invoking principal remains the source of metering, grants, env/KV, lifecycle, and provenance.

## Changes

- Add a `WasmHash` artifact key and principal-scoped registry runtime keys.
- Generate deterministic WASM capsule UUIDs from principal + capsule name + content hash.
- Resolve host-side capsule provenance by UUID to the principal-scoped runtime instance.
- Derive gateway trusted source IDs for session/model capsule replies from installed capsule metadata without changing the public gateway constructor/API.
- Use explicit `v0`/`v1` gateway capsule source-id helpers and read `meta.json` leniently for `wasm_hash`, then derive `v1` IDs from `Capsule.toml` only when metadata lacks a usable hash.
- Scope dispatcher candidate selection for user-facing tool/CLI/describe topics by caller principal view, including LLM provider discovery.
- Gate `cli.v1.command.run.<provider>` as the actual user-invocable capsule command surface.
- Return a bounded CLI cancellation error when a capsule command provider unloads before publishing its command result.
- Scope principal-stamped orchestration/result dispatch to the caller principal's capsule view while preserving admin/global bypass behavior.
- Use an internal principal + capsule + topic readiness key for the in-process capsule topic probe, so hostile same-topic subscribers cannot satisfy session-capsule readiness.
- Keep admin/global listing behavior only for admin callers; ordinary callers get their own capsule inventory.
- Load default and profile principals into separate runtime views while reusing content-addressed artifacts.
- Make install/reload/unload/restart principal-aware.
- Preserve existing public struct construction for `CapsuleContext` and `InstallOptions`.
- Pin the public Rust API diff job to the last known-good nightly so `cargo-public-api` does not float onto a rustc ICE.
- Split dispatcher chain-lock plumbing into a private module so `dispatcher.rs` stays below the source file cap.
- Keep gateway `v1` trusted capsule source IDs caller-only, with `v0` reserved as the package-only compatibility fallback.
- Publish `astrid.v1.capsules_loaded` per principal, include principal identity in entries, and make CLI unload cancellation principal-aware.
- Add an in-flight guard for full capsule reload scheduling so repeated reload requests cannot spawn overlapping all-principal sweeps.
- Materialize cloned capsule installs from the clone source principal, with rollback on materialization failure and copied capsule env removal.
- Preserve the existing `principal_introspection` module/export in `astrid-capsule-install` for public API compatibility while keeping it out of the production clone path.
- Serialize per-principal capsule load/warm operations in the kernel to avoid concurrent WASM load stampedes during live admin flows.
- Split daemon boot loading so default/system capsules gate readiness and profile-principal capsule views warm after the daemon and gateway are ready.
- Warm profile-principal uplink capsules before full profile views so registry-backed HTTP model routes can answer promptly after restart.
- Add/update regressions for same-hash artifact reuse with runtime isolation, scoped dispatch/describe, grant miss behavior, principal-stamped orchestration routing, nonblocking inventory reads, async reload scheduling, per-principal unload/health handling, trusted session/model source-id matching, live per-principal env overlay precedence, LLM provider discovery fan-out scoping, duplicate provider-entry discovery handling, actual CLI run-surface gating, command unload cancellation, and restart-time model registry warm-before-publish behavior.

## Verification

- `cargo test -p astrid-capsule principal_stamped_orchestration_uses_caller_view -- --quiet`
- `cargo test -p astrid-capsule access_enforcement -- --quiet`
- `cargo test -p astrid-capsule llm_describe -- --quiet`
- capsule-registry#24: `cargo test --target aarch64-apple-darwin dedupe -- --quiet`
- capsule-registry#24: `cargo build`
- `bash -n scripts/e2e/runtime-llm-smoke.sh`
- `bash -n scripts/e2e/runtime-process-helpers.sh scripts/e2e/runtime-harness.sh scripts/e2e/runtime-multi-home-smoke.sh`
- `bash -n scripts/e2e/runtime-hostcall-cancel-smoke.sh scripts/e2e/runtime-harness.sh scripts/e2e/runtime-process-helpers.sh scripts/e2e/runtime-gateway-smoke.sh`
- `cargo test -p astrid-integration-tests --test management_api_authz -- --quiet`
- `cargo test -p astrid remove_capsule -- --quiet`
- `cargo test -p astrid check_remote_version -- --quiet`
- `cargo test -p astrid commands::capsule_verb -- --quiet`
- `cargo test -p astrid-capsule-install scan_installed_capsules -- --quiet`
- `cargo test -p astrid-kernel resolve_scope_ -- --quiet`
- `cargo test -p astrid-kernel capsule_topic_probe -- --quiet`
- `cargo test -p astrid-kernel capsule_topic_probe_reflects_loaded_registry_without_capability -- --quiet`
- `cargo check -p astrid-core -p astrid-kernel -p astrid-gateway`
- `cargo clippy -p astrid --all-features -- -D warnings`
- `cargo clippy -p astrid-capsule --all-features -- -D warnings`
- `cargo clippy -p astrid-core -p astrid-kernel -p astrid-gateway --all-features -- -D warnings`
- `cargo check -p astrid-capsule -p astrid-capsule-install -p astrid-kernel`
- `cargo clippy -p astrid-capsule -p astrid-capsule-install -p astrid-kernel --all-features -- -D warnings`
- `cargo test -p astrid-capsule get_config_tests -- --quiet`
- `cargo test -p astrid-capsule ipc_recv_future_drop_leaves_host_state_untouched -- --quiet`
- `cargo test -p astrid-kernel kernel_router -- --quiet`
- `cargo test -p astrid-capsule dispatcher -- --quiet`
- `cargo test -p astrid-gateway routes::capsule_sources -- --quiet`
- `cargo test -p astrid-gateway wrong_capsule_source_reply -- --quiet`
- `cargo test -p astrid-kernel capsules_loaded -- --quiet`
- `cargo test -p astrid-kernel full_reload_guard_coalesces_until_finished -- --quiet`
- `cargo test -p astrid commands::capsule_verb -- --quiet`
- `cargo clippy -p astrid-kernel --all-features -- -D warnings`
- `cargo check -p astrid-daemon -p astrid-kernel`
- `cargo check -p astrid-daemon -p astrid-kernel -p astrid-gateway -p astrid-core`
- `cargo clippy -p astrid-daemon -p astrid-kernel --all-features -- -D warnings`
- `cargo clippy -p astrid-daemon -p astrid-kernel -p astrid-gateway -p astrid-core --all-features -- -D warnings`
- `cargo clippy -p astrid --all-features -- -D warnings`
- `cargo clippy -p astrid-gateway --all-features -- -D warnings`
- `cargo test -p astrid-gateway registry_round_trip_warms -- --quiet`
- `cargo test -p astrid-gateway session_mgmt_gate_warms -- --quiet`
- `cargo test -p astrid-gateway routes::models -- --quiet`
- `cargo test -p astrid-kernel capsule_topic_probe -- --quiet`
- `cargo test -p astrid-kernel agent_create_clone -- --quiet`
- `cargo test -p astrid-kernel agent_create_with_inherit_copies_from_source -- --quiet`
- `cargo test -p astrid-kernel agent_create_without_inherit_copies_nothing -- --quiet`
- `cargo test -p astrid-capsule-install -- --quiet`
- `cargo clippy -p astrid-kernel --all-features -- -D warnings`
- `cargo check -p astrid-daemon -p astrid-kernel`
- `cargo clippy -p astrid-daemon -p astrid-kernel --all-features -- -D warnings`
- `cargo clippy -p astrid-capsule-install --all-features -- -D warnings`
- `cargo fmt --check`
- `cargo metadata --locked --no-deps --format-version 1`
- `git diff --check origin/main...HEAD`
- Local mirror of the PR file-size cap against `origin/main...HEAD`

## Checklist

- [x] Linked to an issue
- [x] CHANGELOG.md updated
- [x] Regression tests added


joshuajbouw:
Reviewer focus for the latest changes:

- Can `astrid capsule run` incorrectly treat a forged or malformed `astrid.v1.capsules_loaded` frame as provider unload?
- Does switching command waiting from `read_until_topic()` to raw-frame scanning preserve behavior for command results and broadcasts?
- If command result and unload broadcast race, does the command result still win when appropriate?
- Is `cli.v1.command.run.<provider>` the correct user-invocable capsule command surface to gate?
- Does this preserve principal-scoped capsule routing without default/admin bleed?

PR #1093, opened 2026-07-01
feat(ipc)!: keepalive-driven inactivity timeout for gateway↔kernel requests (504 on timeout, typed errors)

### LINKED ISSUE #1092: Gateway↔kernel IPC uses a 15s total timeout, turning slow-but-metered ops into misleading HTTP 500s
The gateway→kernel IPC client (`astrid-uplink::kernel_client` / the gateway's `BusKernelClient`) waits for a kernel response with a blanket **15s total deadline** (`DEFAULT_TIMEOUT`). Heavy ops — chiefly `InstallCapsule`, whose kernel handler unpacks the archive and then `ensure_principal_loaded` (loads the capsule + runs its `#[install]` lifecycle hook = instantiate/execute WASM) — can legitimately exceed 15s under load. When they do, the read times out and the gateway maps the transport error to a generic **HTTP 500** (all *real* install failures already map to 403, so a 500 specifically means a transient timeout, not a server bug).

This shows up as intermittent runtime-e2e flakiness (`operator .capsule install expected HTTP 200, got 500`), and mislabels a timeout as an internal error.

The 15s was never the guard against runaway capsules — that is already enforced by fuel (`FuelLedger`), the wasmtime epoch-interrupt deadline, and the memory ledgers. So the IPC timeout has no reason to be lean.

### Proposed fix

Make the IPC timeout an **inactivity** timeout driven by kernel keepalives, not a total deadline:

- The kernel emits a `KernelResponse::Working` keepalive on the request's response topic every ~5s while a handler runs (uniform across all requests; a fast op finishes before the first tick and emits none).
- The uplink resets its inactivity window on each `Working` (swallowed, never surfaced to the client) and returns on the terminal response, so a slow-but-live handler is not mistaken for a dead one.
- An absolute `MAX_TOTAL` (~10 min) ceiling bounds a kernel that keeps pinging forever.
- A genuine timeout is mapped to **504 Gateway Timeout** (honest, retryable), not 500.

This composes cleanly toward SSE-streamed install progress later (the `Working` frames are progress events); a full async 202+poll long-running-operation API remains a larger, separate option.

joshuajbouw:
## Linked Issue

Closes #1092

## Summary

The gateway↔kernel IPC client waited for a kernel response with a blanket **15s total deadline**. Heavy ops — chiefly `InstallCapsule` (unpack archive → `ensure_principal_loaded` = load the capsule + run its `#[install]` lifecycle hook) — can legitimately exceed 15s under load, and when they did the read timed out and the gateway surfaced a generic **HTTP 500** (every *real* install failure already maps to 403, so a 500 specifically meant a transient timeout). That is the intermittent runtime-e2e flake (`operator .capsule install expected HTTP 200, got 500`), and it mislabels a timeout as a server error.

The 15s was never the guard against runaway capsules — fuel (`FuelLedger`), the wasmtime epoch-interrupt deadline, and the memory ledgers already bound execution — so the IPC timeout has no reason to be lean.

This makes the timeout an **inactivity** timeout driven by kernel keepalives, plus an honest status mapping:

- The kernel emits a new `KernelResponse::Working` keepalive on the request's response topic every **5s** while a handler runs — **uniform across every request**, via an RAII pinger dropped before the terminal publish. A fast op finishes before the first tick and emits **zero** keepalives (no overhead on the common path); only genuinely-slow handlers get pinged.
- The uplink (both the socket `KernelClient` and the gateway's bus-direct `BusKernelClient`) resets its inactivity window on each `Working` (swallowed — never surfaced to the HTTP client) and returns on the terminal response, so a slow-but-live handler is no longer mistaken for a dead one.
- An absolute **`MAX_TOTAL` (10 min)** ceiling bounds a kernel that keeps pinging forever.
- A genuine timeout now maps to **`504 Gateway Timeout`** (honest, retryable) instead of `500`; kernel-error → `403` and other transport errors → `500` are unchanged.

The `Working` keepalive reuses the exact `publish_response_value` envelope (kernel session id + `RawJson`), so it is indistinguishable from a terminal frame at the envelope/source level and passes the gateway's existing source-id + correlation-topic filter — the client separates them only by the inner variant. Because each request already uses a unique correlation-UUID response topic, a stray late `Working` lands on a topic nobody is listening to and is harmless.

## Changes

- **`astrid-core`** — new `KernelResponse::Working` keepalive variant (kernel-internal, not WIT) + serde round-trip test.
- **`astrid-kernel`** (`kernel_router`) — `KEEPALIVE_INTERVAL` (5s), a `KeepalivePinger` RAII guard mirroring `EpochTickerGuard`, and a single-sourced `publish_response_value` envelope helper shared by the terminal response and the keepalive; wired uniformly into `handle_request`, dropped before each terminal publish.
- **`astrid-uplink`** (`kernel_client`) — `request()` becomes an inactivity loop that swallows `Working` and resets the window; a downcastable `RequestTimeout` marker + `is_timeout()` so the route layer can detect a timeout; a `MAX_TOTAL` (10 min) ceiling; a `#[cfg(test)]` loopback constructor.
- **`astrid-gateway`** — `BusKernelClient` (the actual install path) rewritten to be keepalive-aware with the same inactivity/`MAX_TOTAL`/`RequestTimeout` semantics; `GatewayError::Timeout` → `504`; `daemon_internal` maps a timeout to 504 and everything else to 500; utoipa 504 doc on `install_capsule`.

No WIT / host-ABI change. No version bump.

## Test Plan

- **Core:** `Working` serializes/round-trips as `{"status":"Working"}`.
- **Uplink (loopback socket):** `Working` frames are swallowed and the terminal returns; keepalives spaced under the window extend past the old 15s *total*; total silence past the window yields the typed timeout; an unending keepalive stream is bounded by `MAX_TOTAL`.
- **Kernel:** a handler slower than the interval emits ≥1 `Working` then stops once the guard drops.
- **Gateway:** `Working` keepalives are swallowed and the terminal returns; a timeout is downcastable and maps to **504**; a kernel `Error` maps to **403** (not timeout); status-code mappings are pinned.
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-features -- -D warnings`, and `cargo test --workspace` pass (verified independently on `astrid-core`/`astrid-uplink`/`astrid-gateway`/`astrid-kernel`).
- End-to-end confirmation (a real slow install over the bus no longer 500-ing) is exercised by the runtime-e2e sweep in CI.


joshuajbouw:
Thanks — all of the Copilot/Gemini feedback was on point; addressed in `5d92cdf`, plus moved the kernel-request client paths off `anyhow` to a typed error:

**Typed errors (no `anyhow` on the request paths).** `KernelClient::request` / `BusKernelClient::request` now return a `thiserror` `KernelClientError { Timeout { kind: Inactivity|Ceiling }, ConnectionLost(#[source] ReadError), BusClosed, Build(#[source] …), Deserialize }` instead of `anyhow::Result` + an `anyhow!(RequestTimeout)` marker + a downcast `is_timeout()`. The gateway now matches `KernelClientError::Timeout` directly for the 504 (else 500; kernel `Error` still 403). This also preserves the underlying transport error as a `#[source]` (fixing the `e.to_string()` chain-loss finding).

**Review fixes:**
- **Hard `MAX_TOTAL` ceiling** — both clients now cap each read at `min(inactivity_remaining, MAX_TOTAL - elapsed)`, so the overall wait can no longer overshoot by a full inactivity window.
- **`tokio::time::Instant`** for `started` in both clients (consistent with the deadline math; works under virtual time).
- **`daemon_internal` de-duplicated** — the bus-client copies collapse into one `pub(crate) daemon_kernel_error` in `routes/mod.rs`.
- **Doc wording** — a fast op still spawns+aborts the guard task, so "no overhead" is now "emits zero keepalive frames / no extra bus traffic".
- **CHANGELOG** — split the `merge=union`-concatenated bullet back into two (this entry closes #1092; the `DeviceKeyId`/#1047 entry restored verbatim).

**Deliberately scoped as follow-ups** (a separate `anyhow`→typed cleanup): `AdminClient::request` still returns `anyhow::Result`, and `KernelClientError::Build` keeps an anyhow `#[source]` only because it wraps `SocketClient::send_message`, which is shared uplink infra still returning `anyhow` — typing that touches other callers and belongs in its own PR.

`cargo fmt`/`clippy --workspace --all-features -D warnings`/`test --workspace` green.

Issue #1168, opened 2026-07-08
Copy-on-write workspace overlay is bypassed by spawned processes


joshuajbouw:
The workspace copy-on-write is an in-process `OverlayVfs`: only file operations routed through the crate observe the overlay's upper layer. A capsule's `host_process` spawn (e.g. `cargo`) calls the OS directly, bypasses the crate, and reads the pristine lower layer — so it operates on stale/absent files and never sees the agent's writes. The overlay's write-isolation is therefore illusory for anything that spawns a process, which is the common case for build/test tooling.

Real copy-on-write must live at the OS/filesystem layer — overlayfs on Linux, APFS `clonefile` on macOS, fail-closed to no-CoW where unsupported — so the fs host and spawned processes share one merged filesystem, with a capability-gated, audited promote/rollback gate for the safety net.

Issue #1167, opened 2026-07-08
Git-managed workspaces incorrectly use the copy-on-write overlay


joshuajbouw:
The per-capsule workspace VFS wraps `workspace_root` in an in-process `OverlayVfs` **unconditionally** (`crates/astrid-capsule/src/engine/wasm/mod.rs`). For a workspace that is under — or contains — git, this is wrong: the agent's writes divert into an ephemeral tempdir upper layer that spawned processes (`cargo`, etc.) and the user never see, while git — which already provides versioning and rollback — is bypassed, and the changes are invisible in `git status` / `git diff`.

A git-managed workspace should write **directly** to the real workspace (git is the rollback and the visibility), reserving copy-on-write for non-git workspaces.

Issue #668, opened 2026-04-23
feat(capsule,kernel): Layer 4 — principal-scoped AllowanceStore, CapabilityStore, OverlayVfs, connection counter (#653)


joshuajbouw:
## Summary

Sub-issue of #653 — **Layer 4** of the production multi-tenancy work. Scope the currently-flat shared-state stores (`AllowanceStore`, `CapabilityStore`), the shared `OverlayVfs`, and the single-counter connection tracker by [`PrincipalId`](../astrid-core/src/principal.rs). These are the four pieces of kernel-level state that today leak across principal boundaries despite Layer 1–3 having already re-scoped everything *inside* `invoke_interceptor`. This layer finishes the **data-structure** isolation that Layer 1 started on the VFS/KV/secret side.

Layer 3 (#666, merged PR #667) shipped per-invocation quota enforcement and stands as the template for this layer's `invoke_interceptor`-time threading: each host function already has `effective_principal()` and `effective_profile()` on `HostState`, and the capsule load context already carries a kernel-scoped shared store (`PrincipalProfileCache`) for principal-keyed lookups. Layer 4 follows that pattern for the approval and capability surfaces.

## Scope

**In:**

- **Gap 12 — `AllowanceStore` principal scoping.** Add a required `principal: PrincipalId` field on `Allowance` (construction-time — *not* `Option`, we want the compiler to flag every call site). Rekey the in-memory `HashMap` from `AllowanceId` → `Allowance` to `(PrincipalId, AllowanceId)` → `Allowance` (or equivalent two-level map). Thread a `&PrincipalId` parameter through `find_matching`, `find_matching_and_consume`, `consume_use`, `export_session_allowances`, `export_workspace_allowances`, and `clear_session_allowances`. All lookups must filter by principal; Agent A's approval never benefits Agent B.
[...]

PR #1170, opened 2026-07-08
feat(capsule): real OS-level copy-on-write for non-git workspaces

joshuajbouw:
## Linked Issue

Closes #1168

## Summary

Replaces the in-process copy-on-write overlay — which spawned processes bypass — with **real OS-level copy-on-write** for non-git workspaces, so the fs host and a spawned `cargo` share one merged filesystem. A `WorkspaceCow` backend abstracts the platform: APFS `clonefile` on macOS, overlayfs on Linux, and a fail-closed no-CoW baseline everywhere else. Changes are committed/discarded through a capability-gated, audited, quiescence-interlocked promote/rollback gate.

Stacked on the git-skip change (#1167) — this PR is based on `fix/git-workspace-skips-cow` and touches only the non-git branch.

## Changes

- New `crates/astrid-vfs/src/workspace_cow/` — the `WorkspaceCow` trait + APFS / overlayfs / no-CoW backends + a `detect`/`prepare` factory that **fails closed to no-CoW with a warn** on any error (never a silently-faked or temp-scattered CoW).
- Wiring (`astrid-capsule`): the non-git workspace uses the backend's `merged_path` as the VFS root, the spawn writable-root, and the spawn cwd, so fs-host writes and spawned processes resolve to one tree. The CoW upper is masked from spawned children in the sandbox (`astrid-workspace/src/sandbox`) so a child cannot write it directly and bypass the gate.
- `KernelRequest::PromoteWorkspace` / `RollbackWorkspace` — capability-gated (`self:workspace:promote` / `rollback`), principal-scoped, audited, and interlocked (refuses while an invocation or spawned child is live) so a commit cannot swap the tree out from under a running process.
- Removed the now-dead `HostState.overlay_vfs` / `upper_dir` fields.

## Test Plan

- macOS runtime (real Seatbelt): APFS clone → isolate → promote → rollback; the sandbox mask blocks a child writing the promote target even under a writable `/var/folders`; a `cwd=merged` **relative** write is seen by the fs host (the process-bypass fix); prepare-failure fails closed to no-CoW.
- Linux overlayfs: compile-checked only (guarded `#[cfg(target_os = "linux")]`, `#[ignore]`d integration test). Runtime validation on CI, plus the opaque-directory whiteout and userns-mount refinements, are a tracked follow-up — not runtime-claimed here.
- `cargo check` / `clippy --all-features --all-targets -D warnings` / `test` / `fmt` green (workspace, `--locked`, toolchain 1.95).

## Notes

Making the workspace per-principal (#668) and the Linux overlayfs runtime validation/refinements are tracked follow-ups.


joshuajbouw:
Superseded by #1171. GitHub auto-closed this PR when its base branch (`fix/git-workspace-skips-cow`, #1169) was deleted on merge, and an auto-closed PR can't be reopened once its base is gone. #1171 is the same branch rebased onto `main` with the review findings here folded in.

PR #1213, opened 2026-07-13
feat: add scoped product control credentials

### LINKED ISSUE #1212: Introduce scoped control credentials for product uplinks
## Goal

Allow a product client to call explicitly permitted local runtime-control operations without reading Astrid session tokens, principal private keys, runtime signing keys, or audit storage directly.

## Acceptance criteria

- runtime verifies a scoped, revocable product control credential before privileged control calls
- credential is bound to an authenticated principal and optional device scope; no legacy anonymous fallback for privileged calls
- operation allowlists prevent a product client from becoming arbitrary runtime admin authority
- the credential path has no dependency on runtime key-file reads outside the runtime-owned boundary
- tests cover absent, malformed, expired/revoked, wrong-principal, and cross-principal credentials
- existing Astrid CLI authentication and the public `astrid.v1` contract remain compatible

## Non-goals

- no AOS HTTP server or gateway migration in this issue
- no runtime bundle/update implementation
- no direct access to kernel handles, EventBus, AuditLog, or gateway signing keys

joshuajbouw:
## Linked Issue

Closes #1212

## Summary

Add the verifier and grant model required for a product-owned control credential, without giving the product the daemon session token or a principal private key.

## Changes

- add an opaque 256-bit credential with redacted debug output and stable wire encoding
- let the product launcher generate and retain the secret while the runtime registers only its SHA-256 verifier
- bind grants to a principal, optional device, explicit closed read-only operation allowlist, expiry, and revocation state
- fail closed for malformed, unknown, expired, revoked, cross-principal, cross-device, and out-of-scope use
- add focused regression coverage and an Unreleased changelog entry

## Verification

- `cargo test -p astrid-core --quiet`
- `cargo clippy -p astrid-core --all-features -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,15 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
 
 ## [Unreleased]
 
+### Added
+
+- **Scoped product-control credentials.** The runtime now has a distinct
+  credential primitive for product clients: a random opaque secret is stored
+  only as a runtime-side hash and is bound to a principal, optional device,
+  explicit read-only operation allowlist, expiry, and revocation state. It
+  cannot fall back to the daemon session token or an anonymous control path.
+  Closes #1212.
+
 ### Changed
 
 - **CLI runtime-control requests now use the shared typed uplink client.**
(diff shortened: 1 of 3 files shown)


joshuajbouw:
Closing without merge. Unicity AOS is a trusted distribution on top of Astrid Runtime and will use the established local operator socket and credentials for its bundled runtime. A separate restricted control credential is deferred until a future component creates a distinct trust domain; it is not a prerequisite for the AOS CLI or product composition.

Issue #1237, opened 2026-07-14
Apply device attenuation to every kernel authority view


joshuajbouw:
## Problem

Primary kernel request checks apply the authenticating device scope, but several secondary authority decisions rebuild principal authority without that device context. A scoped device can therefore receive a broader capsule, principal, or group view than its effective authority permits. A scoped pair request with bare `allow = ["*"]` is also equivalent to unattenuated authority and must use the full-scope privilege gate.

## Scope

- use one typed device-scope resolver for primary and secondary checks
- apply device attenuation to capsule inventory visibility
- apply device attenuation to global principal and group visibility
- require effective `self:auth:pair:admin` for a scoped pair request containing bare `*`
- preserve the `self:*` use-only preset
- deny malformed, unknown, or revoked device IDs
- preserve current no-device and `Full` behavior

## Non-goals

- no public API, WIT, HTTP, IPC, or serialized wire change
- no product-specific identity or trust path
- no capability-registry activation or persisted-policy migration
- capsule dispatcher and gateway audit-probe changes remain separately scoped

## Acceptance criteria

- an attenuated admin device cannot gain global capsule, principal, or group visibility unless its scope admits the exact global capability
- a scoped pair request with bare `*` cannot bypass the pair-admin requirement
- an issuer's own device denies constrain every child scope
- focused kernel regressions cover no-device, Full, scoped, malformed, unknown, and revoked device cases
- formatting, kernel tests, clippy, and repository CI pass

Tracks #1228 and astrid-runtime/rfcs#36.

PR #1242, opened 2026-07-14
fix(gateway): enforce live audit authority

### LINKED ISSUE #1241: Enforce live device policy on audit views
## Problem

The gateway currently infers audit-firehose authority from an agent-list response. That approximation does not evaluate the authenticated device scope, group/revoke precedence, profile enablement, or device revocation through the kernel's actual capability engine. A long-lived SSE connection can also retain its initial visibility after authority is narrowed.

## Required behavior

- evaluate `audit:read_all` through one generic kernel-owned principal/device/capability function
- keep the existing gateway state, default runner, and route-builder APIs source-compatible
- default callers without an in-process evaluator to self-only audit visibility
- evaluate historical audit records after the asynchronous log read
- re-evaluate open SSE streams for each audit event so revocation and narrowing take effect without reconnect
- reject malformed, unknown, revoked, disabled, group-revoked, and device-scope-denied authority
- keep the evaluator borrowed and boolean-only; do not expose signing/runtime keys or caller-selected identities

## Verification

- no-device, `Full`, scoped allow/deny, malformed, unknown, revoked, disabled, and group/revoke policy cases
- historical narrowing suppresses another principal's records while preserving the caller's records
- a real HTTP SSE stream starts with cross-principal visibility, revokes the authenticating device through the kernel, then retains only the caller's own events
- existing public gateway entrypoints remain compatible

Depends on #1237 and the device/cache guarantees in PR #1238.

joshuajbouw:
## Linked Issue

Closes #1241

Refs #1237 and astrid-runtime/rfcs#36.

## Summary

Replaces gateway-side audit authority inference with the kernel's live principal/device capability evaluator. Historical and streaming audit views now narrow through the same runtime policy, with no AOS-specific role, key exposure or caller-selected identity.

## Changes

- add an additive borrowed `CapabilityCheck` constructor while preserving the existing constructor and owned errors
- expose one doc-hidden boolean kernel evaluator for a principal, optional device and capability
- inject the in-process evaluator through the gateway runner and an additive embedded-router builder; the default router remains self-only
- evaluate historical audit authority after the asynchronous log read and at every pagination record boundary
- re-evaluate `audit:read_all` for every live SSE event
- preserve self visibility after malformed, unknown, revoked, disabled, group-revoked or scope-denied authority
- keep raw audit cursor positions stable while live principal narrowing filters same-second batches
- bind new cursors to immutable audit-entry identity and effective visibility scope so same-second appends cannot shift the page boundary and widening or a principal change requires a restart
- reject unsafe scope changes before visible, hidden and cursor-skipped rows instead of returning a partial page or false end-of-results
- accept legacy timestamp/raw-offset cursors only for the unambiguous default self view
- suppress continuation cursors at true physical EOF while retaining a scope-bound cursor when hidden rows make a later widening unsafe
- pin the gateway's live `audit:read_all` enforcement constant and require it to resolve through capability-registry revision 1
- retain the pair-device terminal audit distinctions already present on `main`
- add a race-free pre-bound listener entrypoint for the real HTTP SSE regression
- expose the same live evaluator boundary to trusted in-process router embeddings without exposing the internal probe type
- document the connect-info make-service requirement for both public router builders so unauthenticated redeem throttling keeps the real peer address, and make the default builder's self-only audit policy explicit
- document the additive runner/router migration path for co-located callers that need audit firehose visibility
- accept hyphen-leading opaque invite tokens in both CLI redeem and revoke paths so every base64url token the runtime issues remains usable

## Verification

- exact signed head: `bf3eb251164f1c20e47fb9f3e841677d9efecc02`
- all 20 commits above `d85fb10663c21d08fc43d39d939cfbb5a023cafd` have good Joshua committer signatures
- `git diff --check`
- `cargo doc -p astrid-gateway --no-deps`
- `cargo test -p astrid-gateway` — 160 unit, 7 CORS, 10 models and 18 router tests pass
- `cargo test -p astrid-kernel runtime_policy_tests` — 8 tests pass
- focused audit pagination suite — 17 tests pass, including same-second insertion, missing-anchor and pre-anchor authority-widening regressions
- focused registry enforcement assertion — 1 test passes
- `cargo clippy -p astrid-gateway -p astrid-kernel --all-targets -- -D warnings`
- all 12 production-root CLI parser tests pass, including redeem and revoke with a generated-token shape beginning in `-`
- `cargo clippy -p astrid --all-targets -- -D warnings`
- the replacement GitHub matrix is the execution gate for the combined workspace

## Compatibility

The public HTTP shape, WIT contracts, persisted authority, gateway state and existing runner entrypoints remain unchanged. The new kernel evaluator is boolean and doc-hidden. The internal typed gateway probe remains private; trusted in-process embedders receive an additive generic callback builder with the same fail-closed policy boundary.

When #1244 is refreshed after this PR, its workspace-aware gateway runner must retain this live capability probe across normal, TLS and pre-bound listener paths.

## Checklist

- [x] Linked to an issue
- [x] `CHANGELOG.md` updated under `[Unreleased]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,10 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
 
 ### Fixed
 
+- **Invite commands accept every token the runtime can issue.** `invite redeem`
+  and `invite revoke` now treat a leading hyphen in an opaque base64url token as
+  token data instead of misparsing it as an unknown option.
+
 - **Security Audit now uses a reproducible cargo-audit installation.** CI installs
   cargo-audit 0.22.2 with its published lockfile before invoking the pinned RustSec
   action, keeping the advisory scan compatible with the repository toolchain.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,7 +143,9 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
 - **Audit views now use the live kernel authority evaluator.** Historical and
   streaming firehose access carries the authenticated device scope; revoked,
   disabled, malformed, unknown, or narrowed credentials fall back to the
-  caller's own records. Closes #1241.
+  caller's own records. Historical continuation cursors bind the immutable
+  audit-entry identity and visibility scope, preventing same-second appends
+  from shifting the page boundary and rejecting unsafe widening. Closes #1241.
 
 - **Device attenuation now applies to every kernel authority view.** Capsule,
   agent, and group inventory checks use the authenticating device scope, and a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,6 +140,10 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
   identities fail closed. `capsule:list` controls global describe visibility,
   while exact `capsule:access:any` controls unrestricted execution. Closes
   #1239.
+- **Audit views now use the live kernel authority evaluator.** Historical and
+  streaming firehose access carries the authenticated device scope; revoked,
+  disabled, malformed, unknown, or narrowed credentials fall back to the
+  caller's own records. Closes #1241.
 
 - **Device attenuation now applies to every kernel authority view.** Capsule,
   agent, and group inventory checks use the authenticating device scope, and a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,7 +149,11 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
   disabled, malformed, unknown, or narrowed credentials fall back to the
   caller's own records. Historical continuation cursors bind the immutable
   audit-entry identity and visibility scope, preventing same-second appends
-  from shifting the page boundary and rejecting unsafe widening. Closes #1241.
+  from shifting the page boundary and rejecting unsafe widening. Co-located
+  callers that need firehose visibility must now pass the kernel evaluator via
+  `astrid_gateway::run_with_capability_probe(...)` or
+  `routes::build_with_capability_probe(...)`; the legacy default runner and
+  router remain self-only. Closes #1241.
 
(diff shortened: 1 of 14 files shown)


joshuajbouw:
@copilot review exact head 01d9e494ceca7a9b2965fae7ddfa5edddbba7608

joshuajbouw:
@copilot review exact head 4b1d30bc2a07fdc30c554da161b8096ad3096ea8

joshuajbouw:
@copilot review exact head 05842ae11bc295c94690c778e608ac6f006a0720

joshuajbouw:
@copilot review exact head 8f168b49540c3e535351122221295105a15f82b2

joshuajbouw:
@copilot review exact head edde792268d5d3b7f2588de4fb3557d9ecd5de56

joshuajbouw:
@copilot review exact head ad8f9a7c021a03a53848fe503d0b73079e8ea9a9

joshuajbouw:
@copilot review exact head 2109cd44c26f266eb737c867a9170b30c869dadd

joshuajbouw:
@copilot review exact head 92bd7a083ee699b9545e81dfd880c1d7bc2c0865

joshuajbouw:
@copilot review exact head b477f65b5aefde3e898fede3f8965ca2cadf4f04

joshuajbouw:
@copilot review exact head 39720da99b88957fb5cff0115e598eac2ab618f7

joshuajbouw:
@copilot review exact head 0a38bfee750df4daee152d8bb2e5882ce1d4af72

joshuajbouw:
@copilot review exact head 1f8f5a5ef8130604aff12f0c795132871cca8f3b

joshuajbouw:
@copilot review exact head bf3eb251164f1c20e47fb9f3e841677d9efecc02

PR #1293, opened 2026-07-20
fix(release): restore channel workflow startup

### LINKED ISSUE #1292: fix(release): restore channel workflow startup after stable publication split
## Summary

Channel promotion fails at workflow startup before any job is created after stable crates publication was extracted into a reusable workflow.

The first observed dev attempt was run 29741800581. It ended with `startup_failure`, no jobs, and did not modify the signed dev pointer (which remains generation 3 on v0.10.2).

## Confirmed cause

GitHub's run annotation reports that the nested `publish` job requests `actions: read`, while the caller only delegates `actions: none`. GitHub resolves the reusable workflow before evaluating its stable-only job condition, so this invalid permission boundary blocks dev even though dev would skip crates.io publication.

The reusable workflow also references `secrets.CARGO_REGISTRY_TOKEN`. The interface should explicitly declare and map that optional named secret instead of broadly inheriting caller secrets. The actual publishing job remains protected by the `release` environment; its environment secret is the credential authority.

## Acceptance criteria

- delegate only `actions: read` and `contents: read` to the stable publication call
- explicitly declare and map the optional named crates.io token without broad secret inheritance
- retain `release` environment approval and stable-only crates publication
- add workflow-contract regressions for both boundaries
- validate both workflows with `actionlint`
- prove GitHub can plan the workflow on a non-main ref without mutating a release channel
- successfully promote the already published v0.10.4 candidate to dev generation 4 without invoking crates.io publication

GitHub documentation: https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#using-inputs-and-secrets-in-a-reusable-workflow

joshuajbouw:
## Summary

Restore GitHub's ability to plan the channel-promotion workflow without expanding its publication authority.

## Linked Issue

Closes #1292

## Changes

- delegate the least-privilege `actions: read` and `contents: read` permissions required by the reusable stable-publication job
- declare and explicitly map the optional crates.io token while retaining the protected `release` environment as credential authority
- extend the workflow contract test so both reusable-workflow boundaries remain valid

## Release safety

Dev and nightly still skip the crates.io publication job. Stable publication remains main-only, waits for the authenticated dev candidate, and requires the protected `release` environment.

GitHub planned both jobs and skipped them on the non-main preflight, proving the workflow now compiles without mutating a channel: https://github.com/astrid-runtime/astrid/actions/runs/29742629222

## Verification

- `actionlint .github/workflows/promote-channel.yml .github/workflows/publish-stable-crates.yml`
- `bash scripts/test_channel_workflow_contract.sh`
- `bash -n scripts/test_channel_workflow_contract.sh`
- `git diff --check`


joshuajbouw:
@copilot review

PR #1379, opened 2026-07-25
fix(kernel): isolate management rate limits

### LINKED ISSUE #1378: fix(kernel): isolate management rate limits after authorization
## Problem

The kernel management router charges the global per-method rate-limit bucket
before capability and device-scope authorization.

For `Shutdown`, whose limit is one request per minute, a caller without
`system:shutdown` can consume the only slot. An authorized operator's immediate
shutdown is then rejected for 60 seconds. The same global keying also allows one
principal to consume another principal's reload, install, or workspace-operation
budget.

This was reproduced by the native Windows lifecycle workflow: a deliberately
restricted principal's denied shutdown was followed by an authorized shutdown,
which received `Rate limited: max 1 Shutdown requests per minute`.

## Required behavior

- Authorization, including device-scope attenuation, happens before an action
  consumes its quota.
- Management quotas are isolated by validated principal and request method.
- A restricted principal or device cannot consume an authorized caller's
  action budget.
- Authorized-but-rate-limited requests are audited as failures, not successful
  management actions.
- Inactive principal/method buckets are retired after their sliding-window
  timestamps expire.

## Acceptance criteria

- A denied shutdown does not signal shutdown or block an immediate authorized
  shutdown.
- One principal exhausting a method does not affect another principal.
- The same principal remains limited, and different methods remain independent.
- Sliding-window boundary, partial expiry, stale-key pruning, and bounded queue
  behavior have deterministic tests.
- The regression is fixed in the kernel itself rather than hidden with a sleep
  in platform CI.

joshuajbouw:
## Linked Issue

Closes #1378

## Summary

Prevent denied or unrelated management callers from consuming another
principal's mutation budget. Management action quotas now apply only after full
principal and device-scope authorization and are keyed by principal plus method.

## Changes

- move quota admission behind the capability/device-scope authorization boundary
- isolate sliding-window buckets by validated principal and request method
- periodically retire expired timestamps and inactive principal/method buckets
- audit authorized-but-rate-limited attempts as failures rather than successful actions
- add deterministic boundary, partial-expiry, cross-principal, bounded-queue, and stale-bucket tests
- add a live router regression proving a denied shutdown cannot block an immediate authorized shutdown, while the authorized principal's second shutdown remains limited

## Verification

- `git diff --check`
- `cargo metadata --locked --no-deps --format-version 1`
- focused kernel tests are included for CI; the local macOS host blocked newly linked dependency build scripts in `dyld` before any crate build script executed, both inside and outside the sandbox, so no local compile verdict is claimed

## Checklist

- [x] Linked to an issue
- [x] CHANGELOG.md updated (entry under `[Unreleased]` — or `[Unreleased]` rolled into a version section for a release PR; not applicable to docs/CI-only changes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,12 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
 
 ### Fixed
 
+- **Kernel management rate limits no longer cross authorization boundaries.**
+  Requests consume a principal-and-method-scoped action budget only after
+  principal and device-scope authorization succeeds. A restricted caller or
+  device can no longer exhaust another authorized caller's shutdown, reload,
+  install, or workspace-operation allowance; rate-limit rejections are audited
+  as failures, and expired principal buckets are retired. Closes #1378.
 - **Fresh Windows capsule installs provision private runtime and principal
   homes.** Capsule signing and authority inspection now create and validate the
   Astrid directory tree before generating the runtime identity, secure the new
--- a/crates/astrid-kernel/src/kernel_router/mod.rs
+++ b/crates/astrid-kernel/src/kernel_router/mod.rs
@@ -11,11 +11,10 @@ mod rate_limit;
 mod response;
 mod visibility;
 
-pub(crate) use rate_limit::rate_limit_for_request;
+pub(crate) use rate_limit::{ManagementRateLimiter, rate_limit_for_request};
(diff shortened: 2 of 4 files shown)


joshuajbouw:
@copilot review

joshuajbouw:
@copilot review

joshuajbouw:
@copilot review

Issue #1197, opened 2026-07-12
Run-loop capsule KV writes land in the neutral store instead of the owning principal's KV


jvsteiner:
> **Repo:** unicity-astrid @ `6ea3489` · **Runtime:** astrid 0.9.4 / sdk 0.7.1
> **Type:** bug · **Component:** `crates/astrid-capsule/src/engine/wasm/`
> **Supersedes:** the earlier `2026-07-12-capsule-scoped-kv.md` proposal — that
> asked for a new capsule-scoped store; this is the smaller, correct fix. A
> `run`-flagged capsule already runs one persistent instance per principal with
> KV bound to that principal, so no new store is needed — the binding just isn't
> reaching the run loop's own execution.

## Summary

A `run`-flagged (run-loop) capsule's background work writes KV to the **neutral,
in-memory fail-closed placeholder** rather than the owning principal's
`{principal}:capsule:{id}` store. The write **succeeds silently** — it is a
misroute, not a failure — so the data is invisible to that principal's tool
calls and lost on reload. Root cause: nothing installs the owner's KV overlay on
the run loop's `HostState` before the guest `run` export executes.

## Impact

For a resident run-loop capsule (e.g. an inbox that receives messages over a
network transport and stores them for the operator's tools to read):

- Inbound events the run loop stores are **unreadable by every tool call** (tools
  resolve `{principal}:capsule:{id}`; the run loop wrote neutral).
- **Nothing the run loop persists survives a reload** — checkpoints, cursors,
  work-state all reset. A relay-backed capsule replays from `since=0` every load.
- No error is surfaced to the capsule, so it looks like data silently vanishes.

This is the whole value proposition of a resident capsule (accumulate state from
background events), so for run-flagged capsules it's effectively load-bearing.

## Repro

1. Install a `run`-flagged capsule for a single principal `P`.
2. From its `run` loop (no tool call in flight), `kv::set("k", v)`.
3. Call any tool as `P` that does `kv::get("k")` → returns empty.
4. Reload the capsule → `k` is gone.

Observed on the live daemon: the store call returns success and logs, but the
value never surfaces to a tool read and does not survive reload.

## Root cause (code path, all in `crates/astrid-capsule/src/engine/wasm/`)

1. **Run loop invokes `run` with no overlay install.** `mod.rs:2334-2359`: the
   spawned task locks `run_store`, resolves the typed `run` func (`:2337`), and
[...]

PR #1380, opened 2026-07-26
fix(capsule): bind run loops to typed owner context

### LINKED ISSUE #1224: Run-loop (#[astrid::run]) capsules receive no per-principal resources: operator env overlay and home:// fs both fail
## Environment

- Astrid daemon 0.9.4 (also traced in current source)
- `astrid-sdk` 0.7.1
- Target: `wasm32-unknown-unknown` (Component Model)
- macOS

## Summary

An autostarted **run-loop** capsule (`#[astrid::run]`, started at daemon boot) receives **none of its per-principal resources**: operator env config from `.config/env/<capsule>.env.json` never loads, and `home://` filesystem reads return `CapabilityDenied`. A **bus-invoked** capsule (carrying an inbound principal) gets both. This makes run-loop capsules that need operator config or a data file unusable through the normal channels.

Manifest `net_*` capabilities are unaffected — a run-loop capsule can `bind_tcp` a loopback listener and make outbound airlock HTTP calls fine. The breakage is specific to per-principal resources (env overlay, `home://` fs — and presumably KV/secrets).

## Symptom

Given a run-loop capsule with:

- `.config/env/<capsule>.env.json` = `{"MY_ARTEFACT_PATH": "home://data/artefact.bin", ...}`
- manifest `fs_read = ["home://data/"]`

At boot the guest observes:

- `env::var("MY_ARTEFACT_PATH")` → `Ok("")` (empty), **not** the configured value.
- `env::var("MY_LISTEN_PORT")` (declared nowhere) → `Err` — so the guest's `unwrap_or(default)` fires, correctly.
- `fs::read("home://data/artefact.bin")` → `CapabilityDenied`.

The same capsule invoked over the bus (carrying the `default` principal) sees the configured env value and reads the file successfully.

## Root cause (traced in source)

1. The run loop is driven directly — grab the dedicated run `Store`, call the guest `run` export — at `crates/astrid-capsule/src/engine/wasm/mod.rs:2307-2372`, **bypassing `invoke_interceptor`**.
2. `HostState.invocation_env_overlay` (the `.config/env/<capsule>.env.json` values) is only populated by `load_invocation_env_overlay(principal, capsule_id)` (`crates/astrid-capsule/src/engine/wasm/mod.rs:1156`) at three principal-gated call sites: bus dispatch (`mod.rs:2716`) and the recv installer (`host_state_invocation.rs:177`, `:247`). The boot `run()` invocation carries no principal, so the overlay stays `None`.
3. Guest `env::var` (the `astrid:sys` `get_config` backend, `crates/astrid-capsule/src/engine/wasm/host/sys.rs:24-58`) then falls through to load-time `self.config`, which `resolve_env` (`crates/astrid-capsule/src/engine/mod.rs:194-280`) built from only the **capsule KV store** and **manifest `[env]` defaults** — both principal-independent. Nothing seeds `.env.json` → KV at load, so the operator's values are absent there.
4. `home://` resolves against the principal's home directory; with no principal there is no home to resolve against, so `fs_read` is denied at the gate.

## Impact

Any run-loop capsule that hosts a long-lived service — e.g. a loopback TCP server built on the new `bind_tcp` host fn — and needs operator-supplied config or a data file cannot obtain them through the standard channels. Today's only workarounds are baking config into code defaults and embedding data with `include_bytes!`.

## Suggested fix

- Resolve the run-loop's owning principal (the installing/owning identity) and call `load_invocation_env_overlay` before the boot `run()` invocation, so `.config/env/<capsule>.env.json` reaches run-loop capsules the same way it reaches bus-invoked ones.
- Give run-loop capsules a principal-scoped `home://` (same owning-principal resolution), so `home://` fs works without embedding data files.

Surfaced while building a run-loop loopback gateway on the new `bind_tcp` host fn — this is the blocker that appears once a real run-loop service is built on it. See companion issue on the empty-string env seeding.

jvsteiner:
## Linked Issue

Closes #1224. Refs #1197 and #1454.

## Summary

Autonomous `#[astrid::run]` capsules entered their run export without an invocation principal, leaving owner KV, home/tmp, secrets, env, profile, budgets, log, and cancellation at the neutral floor. On the authority-scoped runtime architecture from #1487, this port installs the typed principal runtime owner's context before the run task is published. Explicit `SystemResident` runtimes remain neutral.

## Changes

- Recognize autonomous owner context only from typed `RuntimeScope::Principal`; never infer authority from `default`, manifest shape, or first-loader state.
- Require and resolve the owning principal profile for principal-resident run exports.
- Derive owner home, environment, and log paths from the kernel-supplied capsule context.
- Install the already UID-scoped KV, home/tmp, secrets, profile, env, budgets, log, and cancellation context before entering the guest run export.
- Reject `SystemResident` and mismatched capsule RuntimeIds in the owner-context installer.
- Preserve existing unstamped compatibility callers and recv-driven publisher context behavior.

## Verification

- `cargo test -p astrid-capsule -- --quiet` — 633 passed
- `cargo test -p astrid-kernel --lib -- --quiet` — 321 passed
- `cargo clippy -p astrid-capsule --all-features -- -D warnings`
- `git diff --check`
- Regressions prove Alice's run-loop KV survives replacement, default and Bob cannot observe or overwrite it, owner home/tmp/secrets/log/profile/env are installed, supplied-home env isolation holds, and system/wrong-capsule scopes fail closed.
- Independent adversarial review: clean

## Scope

This closes autonomous run-loop owner context. #1454 remains separate: the current recv ABI has no reliable handler-return boundary, so automatically restoring owner context after a publisher-scoped recv would be an approximation rather than a sound lifecycle edge.

## AI / Tool Assistance

Assisted-by: Claude:Opus-4.8

Assisted-by: Codex:GPT-5

Jamie Steiner and Claude developed the original fix and tests. Codex ported and hardened the change against #1487's typed RuntimeId/RuntimeScope architecture, expanded isolation and replacement regressions, and performed adversarial review. Joshua reviewed the resulting changes and validation and authorized the signed commit, retaining responsibility for the design and merge decision.

## Checklist

- [x] Linked to an issue
- [x] CHANGELOG.md updated (entry under `[Unreleased]` — or `[Unreleased]` rolled into a version section for a release PR; not applicable to docs/CI-only changes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
   preserves insertion order, reads legacy indexes, and recovers safely across
   write failures and cancellation before or after durable publication. Closes
   #1478.
+- **Principal-resident WASM run loops now enter `run` with their typed runtime owner's durable KV, home, secrets, env, profile, and host-call budgets.** Autonomous work no longer falls back to `default`, a first loader, or neutral state, and explicit system-resident loops remain authority-neutral. Closes #1224; refs #1197.
 
 - **Executable capsule runtimes are now scoped to immutable principal authority rather than shared by content hash.** Identical verified WASM still reuses compiled Wasmtime artifacts, but each `PrincipalUid` receives separate Stores, component instances, guest memory/globals, run tasks, subscriptions, MCP/native processes, readiness, cancellation, and health generations. Explicit `SystemResident` services remain intentional singletons; principal routes admit only their owner plus kernel events by default; and stale dispatcher, health, source-lookup, and process-handle state cannot cross a replacement generation. Closes #1486.
 
--- a/crates/astrid-capsule/src/context.rs
+++ b/crates/astrid-capsule/src/context.rs
@@ -116,8 +116,10 @@ pub struct CapsuleContext {
     ///
     /// One instance per kernel boot, backing [`WasmEngine::invoke_interceptor`](
     /// crate::engine::wasm::WasmEngine::invoke_interceptor)'s per-invocation
-    /// quota resolution. Tests and single-tenant deployments may leave this
-    /// `None` — the engine falls back to the process-global default profile.
+    /// quota resolution. Unstamped compatibility callers may leave this `None`
+    /// and retain the process-global default profile. A typed principal runtime
+    /// with an autonomous `run` export requires it so owner authority and
+    /// sub-budgets cannot silently fall back to another context.
     pub profile_cache: Option<Arc<PrincipalProfileCache>>,
     /// Shared per-principal overlay VFS registry (Layer 4, issue #668).
     ///
--- a/crates/astrid-capsule/src/engine/wasm/host_state_invocation.rs
+++ b/crates/astrid-capsule/src/engine/wasm/host_state_invocation.rs
@@ -4,6 +4,52 @@
 use super::*;
 
 impl HostState {
+    /// Bind an autonomous principal runtime to its load owner's complete host
+    /// context before entering the guest's `run` export.
+    ///
+    /// Principal runtimes already receive authority-scoped KV, home/tmp mounts,
+    /// secrets, and logs when the Store is constructed. The dispatcher and recv
+    /// paths normally add the remaining invocation fields, but an autonomous
+    /// run export bypasses both. Snapshotting the owner resources here and
+    /// installing the resolved profile/env overlay makes every host call observe
+    /// one coherent owner context from the first instruction onward.
+    ///
+    /// The typed runtime identity is checked here, at the mutation boundary.
+    /// This helper deliberately accepts no principal argument: a default alias,
+    /// first loader, or message publisher can therefore never be substituted
+    /// for the Store's immutable load owner.
+    pub(crate) fn install_run_loop_owner_context(
+        &mut self,
+        runtime_id: &crate::registry::RuntimeId,
+        owner_profile: Arc<astrid_core::profile::PrincipalProfile>,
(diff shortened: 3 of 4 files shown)


jvsteiner:
@joshuajbouw ready for review — run-loop capsules get their per-principal KV/env/secrets overlay before the guest `run` export (#1197, #1224).

joshuajbouw:
Review outcome: technically worth advancing. The runtime, Windows, CodeQL, and CI jobs are green. Please update the PR body to the repository template headings (Linked Issue, Summary, Changes, Verification/Test Plan, Checklist) and add the required changelog entry; those are the remaining failing gates. The stated `run + ipc::recv` publisher-identity semantics should remain explicit or be covered by a follow-up issue.

joshuajbouw:
Adversarial review against current `main` found that this PR's owner-overlay intent is correct, but the current hash-shared runtime identity makes the patch unsafe to merge as scoped.

The kernel deduplicates one mutable live runtime per verified WASM hash and constructs the first runtime under synthetic `default`. Installing KV, secrets, home, environment, profile, and budgets into that Store can therefore expose `default` authority to a capsule installed only for a non-default principal and to every other view of the same hash. “First installer owns it” would remain load-order- and restart-dependent.

Tracked as prerequisite #1486. That issue preserves content-addressed bytes and compilation while separating principal-resident Stores, instances, subscriptions, processes, health, and lifecycle by authority scope. It explicitly covers boot, live load, attach-existing-view, derived-agent spawn, reload/restart, unload/delete, stdio MCP, persistent processes, and system-resident services.

After #1486 lands, this PR should be rebased and retain its core change: install the established `PrincipalResident(owner)` overlays before calling the run export, with a non-default persistence/isolation regression.

PR #1382, opened 2026-07-26
fix(kernel): enumerate run-loop capsule tools via the describe fan-out (#1198)

### LINKED ISSUE #1198: Run-loop capsule tools never appear in tools/list (load-time describe returns empty; fan-out fallback never fires)
> **Repo:** unicity-astrid @ `6ea3489` · **Runtime:** astrid 0.9.4 / sdk 0.7.1
> **Type:** bug · **Component:** `crates/astrid-kernel` + `crates/astrid-capsule`
> **Related:** the dual-role capsule gap (run-loop + tools); companion to #1197
> (same `pool = None` root, different symptom).

## Summary

A `run`-flagged capsule's tools never appear in `tools/list`. The kernel captures
a capsule's tool surface at load via `describe_loaded_capsule` →
`invoke_interceptor("tool_describe")`, but run-loop capsules have `pool = None`,
so `invoke_interceptor` returns `NotSupported`. That maps to `Ok(empty)` (not an
error), which the kernel injects as a **present-but-empty** `tools: []`. The
runtime describe fan-out fallback fires only when tools are **absent** (a capture
error), never for present-empty — so a run-loop capsule that ships a working
`tool.v1.request.describe` responder is never consulted, and its tools stay
invisible to MCP clients.

## Impact

- A run-loop capsule that also exposes tools (a resident mailbox, scheduler,
  queue — anything that needs a persistent loop *and* operator-facing tools)
  can't have those tools discovered: `tools/list` shows nothing for it.
- Call-by-name still works (the broker routes an unlisted
  `tool.v1.execute.<name>`), which forces a CLI workaround that spawns its own
  `astrid mcp serve` per call. In practice that workaround leaked serve
  subprocesses and clogged the daemon's handshake capacity under repetition.
  Native enumeration removes the need for the workaround entirely.

## Root cause (citations against `6ea3489`)

1. **`tools/list` is built by the broker, not the serve.** `list_tools` does
   `round_trip(TOOLS_LIST_TOPIC, …)` (`astrid-cli/src/commands/mcp/server.rs:380`);
   the broker assembles the list from `astrid.v1.capsules_loaded`, which carries
   each capsule's tool surface in `meta.tools`.
2. **The kernel captures that surface at load** via
   `describe_loaded_capsule(capsule)` (`astrid-kernel/src/lib.rs:1474`), which
   calls `capsule.invoke_interceptor("tool_describe", …)`
   (`astrid-capsule/src/tool_discovery.rs:49`).
3. **Run-loop capsules can't invoke interceptors.** `invoke_interceptor` is
   `self.pool.as_ref().ok_or_else(|| NotSupported(…))?`
   (`astrid-capsule/src/engine/wasm/mod.rs:2475`), and run-loop capsules set
   `self.pool = None` — the comment is explicit: *"`self.pool` stays None so
   `invoke_interceptor` reports NotSupported for run-loop capsules"*
   (`mod.rs:2367-2371`). So → `NotSupported`.
4. **`NotSupported` becomes `Ok(empty)`, not `Err`.** `describe_loaded_capsule`
   maps `is_unsupported(&e) → return Ok(Vec::new())` (`tool_discovery.rs`).
5. **Empty is injected as *present*.** The kernel takes the `Ok(tools)` arm and
   calls `inject_tools(meta, [])`, which unconditionally inserts the key:
   `obj.insert("tools", tools)` (`astrid-kernel/src/capsules_loaded.rs:47`). So
   the capsule advertises `meta.tools = []`.
6. **The fan-out fallback fires only on *absent* tools.** `lib.rs:1470`: *"a
   describe (or serialize) failure leaves `tools` absent and the consumer falls
   back to its fan-out for this cycle."* Failures hit the `Err` arms
   (`lib.rs:1508/1512`) that do **not** call `inject_tools`. A run-loop capsule
   returns `Ok(empty)` → present-empty → **no fan-out**.

## Why the existing fan-out doesn't save it

The runtime describe fan-out (`tool.v1.request.describe` →
`tool.v1.response.describe.*`) already exists, and a run-loop capsule can answer
it (ours does — a manual responder over those topics). But because the load-time
capture returns present-empty rather than absent, the consumer treats the capsule
as "has zero tools" and never fans out. The working responder is structurally
bypassed. Empirically confirmed: our capsule's describe responder runs, yet its
tools never enumerate.

## Proposed fix (small; reuses machinery that already exists)

When `describe_loaded_capsule` returns `NotSupported` for a pool-less / run-loop
capsule, **leave `tools` absent rather than injecting `[]`** — so the consumer's
existing fan-out fallback fires and the capsule's own
`tool.v1.request.describe` responder supplies the surface. One conditional in the
load-time capture path; no new protocol; run-loop capsules enumerate through the
path they already implement.

(Alternative, larger: natively support the interceptor path for run-loop
capsules — but that overlaps the broader dual-role fix and is more invasive.)

## Caveats

- The broker capsule's source (the sage-mcp broker) was not read line-by-line;
  the "fan-out on absent, not on present-empty" behavior rests on the kernel's
  documented contract (`lib.rs:1470`) plus the empirical non-enumeration.
- Same `pool = None` root as the dual-role dispatch gap: a run-loop capsule can
  neither be *dispatched to* nor *described* through the interceptor path. If the
  broader dual-role fix lands, this is subsumed; the conditional above is the
  minimal standalone fix.

jvsteiner:
## Linked Issue

Closes #1198.

## Summary

A run-loop capsule's tools could remain invisible. Load-time interceptor discovery is unavailable for pool-less run-loop capsules, but persisted or injected `tools: []` metadata suppressed the existing describe fan-out that should query the capsule's live responder.

## Changes

- Add fan-out-aware discovery status while preserving the existing public `describe_loaded_capsule` signature.
- Leave `tools` absent when the interceptor path is unavailable so `tool.v1.request.describe` fan-out runs.
- Treat `tools` as live-owned metadata and remove legacy or stale tool surfaces before each probe while preserving unrelated opaque metadata.
- Cover every result-mapping branch plus legacy object metadata, absent metadata, and non-object metadata.

## Verification

- Tool-discovery tests: 16 passed
- `cargo test -p astrid-kernel --lib -- --quiet`: 304 passed
- `cargo clippy -p astrid-capsule -p astrid-kernel --all-features -- -D warnings`
- `git diff --cached --check`

## AI / Tool Assistance

Assisted-by: Anthropic Claude: Opus 4.8

Claude assisted Jamie with the original implementation. Codex independently reviewed the live discovery and persisted-metadata behavior, fixed the stale metadata suppression path, added regressions, reconciled the patch onto current `main`, and validated the affected crates. I reviewed the final combined patch and created the GPG-signed, DCO-compliant commit. Jamie and Claude remain credited as co-authors.

## Checklist

- [x] Linked to an issue
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Run-loop capsule tools now appear in `tools/list`.** A pool-less run-loop capsule's tool-describe path returned `NotSupported`, which was captured as a present-but-empty tool set, so the describe fan-out — which fires only on *absent* tools — never consulted the capsule's own `tool.v1.request.describe` responder. The load-time capture now distinguishes "couldn't capture" from "captured, empty" and leaves tools absent for pool-less capsules so the fan-out fires. Closes #1198.
 ### Removed
 
 - **The unused SurrealDB query wrapper and its dormant dependency graph have
--- a/crates/astrid-capsule/src/lib.rs
+++ b/crates/astrid-capsule/src/lib.rs
@@ -51,7 +51,10 @@ pub use memory_ledger::MemoryLedger;
 // `StoreMemoryMeter` is the Wasmtime `ResourceLimiter`; native-only.
 #[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
 pub use memory_ledger::StoreMemoryMeter;
-pub use tool_discovery::{ToolDescriptor, describe_loaded_capsule, tools_missing_execute_route};
+pub use tool_discovery::{
+    ToolDescriptor, describe_loaded_capsule, describe_loaded_capsule_status,
+    tools_missing_execute_route,
+};
 
 /// Test-only access to security boundaries that integration tests must drive
 /// through real operating-system transports.
--- a/crates/astrid-capsule/src/tool_discovery.rs
+++ b/crates/astrid-capsule/src/tool_discovery.rs
@@ -46,18 +46,63 @@ pub struct ToolDescriptor {
 /// Returns an error if the interceptor errors for any reason other than "not
 /// implemented", genuinely denies (a reason other than the unknown-action one),
 /// or returns a payload that is present but not the expected JSON shape.
+///
+/// Preserves the original signature: a pool-less run-loop capsule's *unknown*
+/// surface collapses to an empty vec here. Callers that must distinguish
+/// "unknown" from "empty" — to drive the describe fan-out (#1198) — use
+/// [`describe_loaded_capsule_status`] instead.
 pub async fn describe_loaded_capsule(capsule: &dyn Capsule) -> anyhow::Result<Vec<ToolDescriptor>> {
-    let result = match capsule.invoke_interceptor("tool_describe", &[], None).await {
+    Ok(describe_loaded_capsule_status(capsule)
+        .await?
+        .unwrap_or_default())
+}
+
+/// Like [`describe_loaded_capsule`], but preserves the "surface unknown" signal
+/// the load-time capture needs.
+///
+/// Returns:
+/// - `Ok(Some(tools))` — the describe ran and returned a (possibly empty) tool
+///   surface; the caller injects it into the `capsules_loaded` meta.
+/// - `Ok(None)` — the describe COULD NOT run because this is a pool-less
+///   run-loop capsule (`invoke_interceptor` reports `NotSupported`). The caller
+///   must then leave `tools` ABSENT (not `[]`), so the consumer's describe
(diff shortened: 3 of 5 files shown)


jvsteiner:
@joshuajbouw ready for review — pool-less run-loop capsules now leave tools absent so the describe fan-out fires and their tools enumerate (#1198).

joshuajbouw:
Review outcome: useful fix, but not mergeable as posted. The public Rust API compatibility job fails because `describe_loaded_capsule` changes from `Result<Vec<ToolDescriptor>>` to `Result<Option<Vec<ToolDescriptor>>>`. Please preserve the existing public signature with an additive/internal API, or explicitly handle this as an intentional breaking change under the project’s release policy. The PR template and changelog also need updating.

PR #1383, opened 2026-07-26
refactor(kernel): AgentDelete always reclaims a deleted principal's footprint (#1217)

### LINKED ISSUE #1217: Spawn a locked-down session on demand: derived principal, caller-chosen capabilities, no network egress
### Background (what I'm building)

I've built a capsule that lets agents send each other messages over NOSTR —
basically "one agent hands another a task, a question, or a bug report." Today
those messages just land in an inbox and I read them later.

The next step I want is to *automatically* react to some of them: when a
message comes in, spin up a **short-lived, throwaway Claude session** that does
one small, bounded job (read the message, gather a bit of context, maybe draft
a reply), leaves the result for me to approve, then goes away. It never acts on
its own in my real session.

### Why it has to be locked down

These messages come from outside, so I treat them as untrusted — a hostile
message could try to trick that session into leaking data. So the session needs
to run **locked down**: only the abilities that job needs, and for the most
cautious jobs, **no network access at all**. If the session has the (possibly
malicious) message *and* a network tool, it could be talked into phoning that
data out — so cutting off the network entirely is the safety guarantee I'm
leaning on.

### The request

When my code kicks off one of these sessions, can astrid launch it as a
**different, more restricted principal** than the one that triggered it
(something like `claude-code-triage-<project>` — the exact name doesn't
matter), where **I choose that session's capability set**, and in particular
can **switch its network/egress off entirely**?

Concretely:

- Can `claude-runner` (or whatever spawns a session) start it as a *derived*
  principal that I name at spawn time?
- Are a session's capabilities fixed per principal, or can the caller hand it a
  restricted set when spawning?
- Is "no network at all" something the capability model can express today?

### Why I'm asking

If this already works, then the security of the whole feature is just "let the
kernel enforce the capabilities" and I don't have to build a second sandbox of
my own. If it doesn't work yet, then *that's* the piece to build first — so
mostly I need to know which of those two we're in.

This is the first of two related asks about the same feature (the second is
about *where* the trigger logic that decides when to spawn a session should
live). It's part of the same effort as the run-loop KV durability bug #1197.
Not urgent — I can't start building this until all three pieces land, so this
is scoping the work, not blocking on you.

jvsteiner:
## Linked Issue

Closes #1217.

## Summary

Deleting an agent must retire an authority generation, not merely remove files. This makes deletion fail closed across identity, tokens, allowances, capsule state, signing material, and crash recovery.

## Changes

- Preserve the public AdminRequestKind::AgentDelete shape.
- Install a durable alias reservation before cleanup, including legacy or partially deleted identities without an immutable UID.
- Fence capability-token and allowance creation and authorization while a principal is retiring.
- Purge session and persistent tokens without erasing global revocation or replay tombstones.
- Purge every native KV namespace through the immutable principal UID, including orphaned namespaces.
- Retire live capsule views without unloading shared runtimes.
- Reclaim the principal home, signing key, and secrets.
- Return an error and retain the reservation if reclamation is incomplete. Successful responses retain an empty cleanup_errors array for wire compatibility.
- Prevent recreation while residual authority or state remains.

## Verification

- cargo test -p astrid-kernel agent_delete
- cargo test -p astrid-storage ownership::tests
- cargo test -p astrid-capabilities store::tests
- cargo test -p astrid-approval allowance::store::tests
- cargo clippy for all affected crates with all features and warnings denied
- repository file-size check reproduced locally

Regression coverage includes cleanup failure followed by recreate and retry, legacy alias tombstones, orphan KV reclamation, retirement fences, revocation preservation, and peer-principal isolation.

## AI / Tool Assistance

Codex performed adversarial review, implemented the fail-closed deletion and recovery fixes, added regression coverage, and validated the affected crates. Human-authored DCO and GPG signatures remain on every commit.

## Checklist

- [x] Linked to an issue
- [x] CHANGELOG.md updated
- [ ] I understand every change in this PR and can explain its design, risks, and validation.
- [ ] I reviewed and tested any meaningful tool-generated output included in this PR.
- [x] Every non-bot, non-merge commit has a matching Signed-off-by trailer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
   store and live legacy-SurrealKV migration reader are unchanged. Closes #1448.
 ### Changed
 
-- **`agent delete` now reclaims the deleted principal's full runtime footprint.** Delete closes authz first (unlink + profile removal + cache invalidate), then retires its live capsule views and reclaims capsule KV namespaces, the home tree, signing key (`keys/{principal}.key`), and secrets (`secrets/{principal}/`). Shared runtimes remain available to other principals while principal-scoped work is cancelled before state removal. Reclamation is best-effort; failures are reported in the response's `cleanup_errors`. Replaces the previous "reclamation is an ops concern" leave-behind, and drops the interim `--purge-home` flag. Part of #1217.
+- **`agent delete` now reclaims the deleted principal's full runtime footprint.** Delete fences new token and allowance authority, unlinks authentication, removes the profile, retires live capsule views, purges every immutable-UID KV namespace (including orphaned capsules), and reclaims the home tree, signing key (`keys/{principal}.key`), and secrets (`secrets/{principal}/`). Reclamation fails closed: incomplete cleanup returns an error, retains a durable alias reservation, and is safe to retry without letting a replacement identity inherit residual authority or state. Successful responses retain an empty `cleanup_errors` array for wire compatibility. Shared runtimes remain available to other principals. Replaces the previous "reclamation is an ops concern" leave-behind, and drops the interim `--purge-home` flag. Part of #1217.
 
 ### Added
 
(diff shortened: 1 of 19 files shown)


jvsteiner:
@joshuajbouw ready for review — opt-in `AgentDelete { purge_home }` to reclaim a throwaway principal's footprint (home + global keys/secrets). First of two for #1217.

joshuajbouw:
Review outcome: valuable direction, but hold for changes. CI reports a public API break because adding `purge_home` adds a field to the public `AdminRequestKind::AgentDelete` enum variant. Please make the API compatibility decision explicit and resolve it under the project’s policy. Also please demonstrate the cleanup contract for active principals and make cleanup failure observable rather than only logging a warning. The linked issue, PR template, and changelog gates still need fixing.

PR #1384, opened 2026-07-26
feat(cli): agent spawn — atomic locked-down throwaway session (#1217)

### LINKED ISSUE #1217: Spawn a locked-down session on demand: derived principal, caller-chosen capabilities, no network egress

jvsteiner:
## Linked Issue

Closes #1217 (built on #1383).

## Summary

Adds `astrid agent spawn`: one bounded job under a newly derived, restricted principal, followed by reclaiming teardown. The derived runtime receives only explicitly named WASM capsule installs, capsule-scoped state, user-invocable capsule grants, and outbound endpoints. Nothing else is inherited implicitly.

## Changes

- Add `astrid agent spawn --job ...` with explicit `--load-capsule`, `--allow-capsule`, `--inherit-capsule-state`, and `--allow-egress` controls.
- Make restricted authority an invariant: spawn exposes no arbitrary capability-grant flag and always provisions the built-in `restricted` group.
- Add a dedicated additive `admin.agent.derive` request topic and client method rather than adding a variant to the exhaustive public `AdminRequestKind` enum.
- Reject duplicate or inconsistent capsule selections, malformed egress, residual target state, and host-MCP/native-process capsule installs.
- Materialize only selected capsule installs and copy env, KV, and declared secrets only for explicitly selected capsule namespaces.
- Enforce restricted-principal network and process policy again at host-call time.
- Authenticate the job uplink as the derived principal, auto-deny approval requests, apply a wall-clock timeout, and send cooperative cancellation before teardown.
- Roll failed provisioning back through the ownership deletion guard so concurrent fleet assignment cannot leave ownership pointing at deleted identity state.
- Drain admitted host effects and terminate principal-owned process groups before reclamation, while preserving peer runtimes.
- Reclaim KV before removing the profile needed by the production quota resolver, retaining the retirement fence throughout teardown.
- Tear down through #1383's reclaiming `AgentDelete`; `--keep` remains an explicit debugging escape hatch.

## Verification

- `cargo test -p astrid-kernel --lib -- --quiet`: 309 passed
- `cargo test -p astrid-capabilities --lib -- --quiet`: 78 passed
- Native deletion and restricted derive regressions
- Retirement drain, late-admission, KV/FS/secret denial, and secret-read barrier regressions
- Process-group descendant cancellation with peer continuity
- Strict non-default profile loading and registration-versus-unload serialization
- Wasm portability script
- Exact CI clippy package surface with all targets/features and `-D warnings`
- `git diff --check`

The live spawn-run-teardown proof passed against a fresh restricted principal derived from `gemma-code`: the real React and OpenAI-compatible provider capsule set warmed, the provider returned exactly `PR1384_LIVE_OK`, teardown succeeded, and the principal disappeared from `agent list`. Its profile, home, key, secrets, and isolated test homes were absent afterward.

## AI / Tool Assistance

Codex helped reconstruct and adversarially review the patch on current code, remove capability-escalation paths, preserve public Rust API compatibility, integrate fleet-ownership-safe rollback, harden retirement and teardown ordering, add focused regressions, and execute the live provider-backed spawn-run-teardown proof. I reviewed the final changes, risks, and validation and created each GPG-signed, DCO-compliant commit.

## Checklist

- [x] Linked to an issue
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
   store and live legacy-SurrealKV migration reader are unchanged. Closes #1448.
 ### Changed
 
-- **`agent delete` now reclaims the deleted principal's full runtime footprint.** Delete closes authz first (unlink + profile removal + cache invalidate), then retires its live capsule views and reclaims capsule KV namespaces, the home tree, signing key (`keys/{principal}.key`), and secrets (`secrets/{principal}/`). Shared runtimes remain available to other principals while principal-scoped work is cancelled before state removal. Reclamation is best-effort; failures are reported in the response's `cleanup_errors`. Replaces the previous "reclamation is an ops concern" leave-behind, and drops the interim `--purge-home` flag. Part of #1217.
+- **`agent delete` now reclaims the deleted principal's full runtime footprint.** Delete fences new token and allowance authority, unlinks authentication, removes the profile, retires live capsule views, purges every immutable-UID KV namespace (including orphaned capsules), and reclaims the home tree, signing key (`keys/{principal}.key`), and secrets (`secrets/{principal}/`). Reclamation fails closed: incomplete cleanup returns an error, retains a durable alias reservation, and is safe to retry without letting a replacement identity inherit residual authority or state. Successful responses retain an empty `cleanup_errors` array for wire compatibility. Shared runtimes remain available to other principals. Replaces the previous "reclamation is an ops concern" leave-behind, and drops the interim `--purge-home` flag. Part of #1217.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
   dependencies in `Cargo.lock`. Removing the unused surface eliminates both
   RustSec exceptions and keeps the audit gate strict. The native principal
   store and live legacy-SurrealKV migration reader are unchanged. Closes #1448.
+### Changed
+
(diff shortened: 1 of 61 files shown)


jvsteiner:
@joshuajbouw ready for review — `agent spawn`: atomic locked-down throwaway session composing shipped primitives. Second of two for #1217; **stacked on #1383** (diff includes it until that merges).

joshuajbouw:
Review outcome: not ready to advance; please keep this as draft or rework it. The “locked-down” guarantee is not enforced as an invariant because callers can supply `--group`; the live run/warm-up path is not demonstrated; and teardown depends on best-effort purge. Please land or resolve the purge design first, constrain the security surface, provide live end-to-end evidence, and fix the required PR template/changelog. I have not approved the currently gated workflows.

Issue #1392, opened 2026-07-26
storage: make bulk content ingest read-bound and delta-proportional


joshuajbouw:
## Outcome

Make first ingestion of large principal content read-bound and make repeated
ingestion proportional to the actual delta.

The performance target is a first 1 TB ingest in roughly 3–6 minutes on a host
whose source and destination sustain that rate, with re-ingest completing in
seconds to minutes when little changed. This is a measured target, not a
hardware-independent promise.

This builds on:

- #1386 for compaction and a persistent object index;
- #1388 for a single appender and group durability; and
- #1389 / #1390 for the pinned content grammar and streaming-builder boundary.

Heavy content workloads remain gated on #1386.

## Change detection without weakening provenance

Maintain a disposable cache:

```text
(source scope, path, size, mtime, stable file identity, source epoch)
    -> verified ObjectId
```

The cache is never part of an export and never establishes object identity.
Every referenced `ObjectId` must already exist and pass ordinary verified reads.
A missing, malformed, stale-epoch, or inconsistent entry falls back to reading
and hashing the source.

There are two trust modes:

1. Astrid-owned sources, or platform sources with a trusted monotonic change
   journal, may reuse a prior `ObjectId` without reading bytes when the complete
   change token is unchanged.
2. Untrusted external imports may use stat metadata to prioritize work, but may
   not publish a cache hit as byte identity without re-reading. A process can
   replace bytes while preserving `(path, size, mtime, inode)`; treating that
   tuple as proof would make the resulting root a claim about stale bytes.

Receipts must distinguish `bytes-observed` from `change-token-observed`. If a
deployment chooses backup-tool-style stat trust as an operator policy, that
weaker observation must be explicit rather than represented as verified ingest.

## Bulk transaction and marker protocol

One batch has a random batch ID and a durable marker containing:

- source scope/epoch and requested catalog mutation;
- expected principal root generation and identity;
- chunking/profile versions;
- a digest of the ordered source manifest; and
- state: `preparing` or `root-published`, plus the published root when known.

Protocol:

1. Capture source metadata and open stable read handles where the platform
   permits it. Write and flush the `preparing` marker.
2. Readers feed bounded buffers to parallel chunk/hash workers. A single
   appender checks the in-memory object index before encoding and coalesces new
   frames into megabyte-scale writes. Identity is still recomputed on admission.
3. Build and closure-validate the complete proposed catalog/root without
   publishing it. Detect source mutation and retry affected inputs.
[...]

Issue #1387, opened 2026-07-26
storage: in-process engine reopen after RequiresRecovery


joshuajbouw:
Follow-up from #1377 (see the pre-merge review comment). Any I/O error mid-`persist` poisons the `DurableEngine` and every subsequent operation returns `RequiresRecovery` — correct fail-secure behavior. But the kernel holds the `Arc<dyn KvStore>` from boot with no reopen path, so a transient fault (ENOSPC blip, EIO retry) leaves the entire KV surface — capability store, identity store, all capsule state — dead until a manual daemon restart, even after the condition clears.

Work: on `RequiresRecovery`, drop and reopen the engine in place under the store lock. Recovery already reconciles disk state (tail truncation, closure re-validation), so an in-process reopen is semantically identical to a daemon restart. Bound it (retry budget + backoff) so a persistently failing disk still surfaces as loud errors rather than a reopen loop.

PR #1395, opened 2026-07-27
feat(storage): add crash-recoverable native write staging

### LINKED ISSUE #1394: storage: add crash-recoverable native content staging
## Problem

Writable filesystem providers must not make user-visible `write()` latency wait
for chunking, hashing, object admission, and principal-root publication. They
also cannot acknowledge bytes that disappear when the daemon crashes, publish
two closes out of order, or let a guest-controlled path bypass the verified
content commit path.

## Scope

Implement the shared native staging backend for the macOS, Windows, and Linux
provider adapters tracked by #1392:

- private random-access staging files with write, seek, and truncate semantics;
- a durable close boundary using flushed bytes and a versioned, checksummed
  intent;
- crash recovery before and after the ready-directory transition;
- close-order publication per owner and content name;
- background ingestion through the existing streaming content builder and
  principal-root CAS;
- idempotent recovery if the root commits before staging cleanup; and
- fail-closed handling for redirected sources, malformed markers, changed
  lengths, and unexpected directory entries without deleting acknowledged
  bytes.

The staging directory is engine-private and never a guest-visible mount.
Platform filesystem providers, staged-byte resource leases, bulk-ingest
scheduling, and contiguous `BlobId` adoption remain separate #1392 work.

## Acceptance

- Sealing returns before content addressing begins but after bytes and intent
  are recoverable.
- A sealed pre-rename write is promoted on reopen.
- A post-root/pre-cleanup crash retries to the same identity and root without
  new objects.
- A later close cannot publish before an earlier close for the same owner/name.
- Symlink/reparse and marker-tamper cases preserve bytes and fail closed.
- Native macOS and Windows builds pass, along with workspace tests, clippy, and
  wasm portability checks.

joshuajbouw:
## Linked Issue

Closes #1394

Part of #1392. Builds directly on merged #1390.

## Summary

Adds the shared native write-staging backend required by future writable
filesystem providers on macOS, Windows, and Linux.

A provider writes an ordinary private random-access file and acknowledges the
close after bytes plus a checksummed intent are recoverable. Chunking, object
admission, quota validation, and authoritative principal-root publication run
afterward on a blocking worker through the existing content-store CAS. Guest
writes never mutate the object store directly.

## Changes

- Add `AstridHome::content_staging_path()` at
  `var/content-staging/`. This is engine-private state and is never projected
  into a guest filesystem.
- Add `NativeContentStagingArea`, `StagedContentWriter`, and
  `ReadyStagedContent` as additive native storage APIs.
- Support ordinary random-access writes, seeks, and truncation without
  buffering a complete value in memory.
- Make `seal()` the hosted-provider acknowledgement boundary:
  - flush the native content file;
  - persist a versioned, checksummed intent; and
  - atomically move the staging directory into the ready queue.
- Allocate the publication sequence at close rather than open. A slow older
  handle therefore cannot overwrite a newer close.
- Enforce that an earlier close for the same owner and content name publishes
  before a later close. Unrelated names do not share this dependency.
- Recover a valid sealed write interrupted before its final directory
  transition. Preserve unsealed entries under `quarantine/` for diagnosis.
- Stream ready files through `PrincipalContentStore::put_streaming_with_profile`
  on `spawn_blocking`, retaining the existing identity verification, closure
  validation, quota calculation, and principal-root CAS.
- Hash identities and encode a staged object batch before taking the durable
  engine mutex. The critical section now contains only index/collision reads
  and the coalesced append.
- Write a durable publication marker only after the root commit. Queue scans
  inspect that marker before requiring content or intent, so every cleanup
  crash prefix is reaped without blocking unrelated ready entries.
- Fail closed on symlink/reparse sources, non-canonical queue names, malformed
  or torn intents, changed lengths, and unexpected directory entries without
  deleting acknowledged bytes.
- Extract the migration and staging filesystem operations into one private
  native I/O boundary instead of duplicating atomic-write, directory-sync, and
  quarantine logic.
- Split lifecycle, wire format, recovery mechanics, and adversarial tests into
  separate modules.

## Crash semantics

| Interruption | Recovery |
|---|---|
| Before intent | Bytes were not acknowledged; preserve under quarantine |
| After intent, before ready rename | Validate intent and source, then promote |
| Before principal-root CAS | Ready file remains and can retry |
| After root CAS, before publication marker | Replay is identity- and root-idempotent |
| After marker, during content/intent removal | Marker proves publication; remove the remaining known files |
| After marker removal, before directory removal | Reap the empty canonical ready directory |

## Security boundary

- The provider supplies a host-stamped `StateOwner`; guest payloads do not
  select principals.
- The staging root is private runtime state, not a filesystem capability.
- Source files and every existing ready-directory member are checked as
  non-redirected regular files before ingest or cleanup.
- Cleanup removes only the three known files and the now-empty directory. It
  never recursively follows an unexpected entry.
- The daemon singleton/principal-store lock remains the exclusion boundary for
  two daemons across an upgrade window.

## Deliberately separate

- macOS FSKit, Windows provider, Linux FUSE/9P, and bare-metal adapters;
- staged-byte reservation against the principal's live resource lease;
- bulk-ingest scheduling, change detection, and parallel chunk/hash workers;
- adopting the contiguous staging file as a physical `BlobId`
  representation;
- per-entry corrupt-intent isolation before a writable provider ships;
- in-process engine recovery after transient ENOSPC (#1387), now a
  pre-provider requirement; and
- direct capsule WIT access while interface PRs remain frozen.

## Verification

Current head `0b31cef954c80b2118d2f7b8f43f5f348abac70f` passed:

- `cargo test --workspace -- --quiet` outside the sandbox — complete workspace
  and doctest pass;
- `cargo +1.95 clippy --workspace --all-features --all-targets --locked -- -D warnings`;
- `cargo test -p astrid-storage-engine -p astrid-storage --all-features -- --quiet`
  — 210 tests passed;
- `cargo check -p astrid-storage --target aarch64-pc-windows-msvc`;
- `cargo check -p astrid-storage --target wasm32-unknown-unknown`;
- `cargo fmt --all -- --check`; and
- `git diff --check`.

Regression coverage includes native close-before-ingest behavior, random-access
write semantics, complete store reopen, pre-rename recovery, post-root retry,
same-name close ordering, corrupt-intent preservation, symlink rejection,
every cleanup-operation prefix with an unrelated pending entry, canonical
marker decoding, and proof that batch identity work does not hold the engine
mutex.

## Checklist

- [x] Linked to an issue
- [x] `CHANGELOG.md` updated under `[Unreleased]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,7 +96,15 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
   incrementally in four-MiB coalesced shared-engine batches without per-object
   flushes; the ordinary root transaction revalidates the complete closure,
   flushes objects before the root, and retries root conflicts without rereading
-  the source. Source or sink failure never publishes a partial root.
+  the source. Source or sink failure never publishes a partial root. Hosted
+  writable projections can now stage random-access writes in private native
+  files and acknowledge a close after durable bytes plus a checksummed intent,
+  without waiting for content addressing. Startup promotes sealed pre-rename
+  writes, preserves incomplete entries for diagnosis, and rejects redirected
+  or malformed ready entries without deleting acknowledged bytes. Background
+  publication runs on a blocking worker, enforces close order per owner and
+  content name, and uses a durable post-root marker so a crash between the root
+  CAS and cleanup retries idempotently.
   Full and range reconstruction validate the typed graph and load only
   overlapping chunks. `PrincipalContentStore` publishes named content through
   the same per-principal root CAS and durable object arena as KV, so identical
--- a/crates/astrid-core/src/dirs.rs
+++ b/crates/astrid-core/src/dirs.rs
@@ -475,6 +475,16 @@ impl AstridHome {
         self.var_dir().join("principal-store")
     }
 
+    /// Private native write-staging area for principal content (`var/content-staging/`).
+    ///
+    /// Filesystem providers acknowledge writes from this area before content
+    /// ingestion publishes them into the authoritative principal store. It is
+    /// engine-private state and must never be projected into a guest view.
+    #[must_use]
+    pub fn content_staging_path(&self) -> PathBuf {
+        self.var_dir().join("content-staging")
+    }
+
     /// Root directory for OS-level copy-on-write workspace clones (`cow/`).
     ///
     /// Each non-git capsule workspace gets a per-workspace subdirectory here
--- a/crates/astrid-core/src/dirs_tests.rs
+++ b/crates/astrid-core/src/dirs_tests.rs
@@ -171,6 +171,10 @@ fn test_astrid_home_fhs_paths() {
         home.principal_store_path(),
         PathBuf::from(format!("{r}/var/principal-store"))
     );
+    assert_eq!(
+        home.content_staging_path(),
+        PathBuf::from(format!("{r}/var/content-staging"))
+    );
     assert_eq!(home.run_dir(), PathBuf::from(format!("{r}/run")));
     assert_eq!(
         home.socket_path(),
--- a/crates/astrid-storage/src/lib.rs
+++ b/crates/astrid-storage/src/lib.rs
@@ -76,7 +76,8 @@ pub use secret::{FallbackSecretStore, KeychainSecretStore};
 pub use kv::SurrealKvStore;
 #[cfg(not(target_family = "wasm"))]
 pub use principal_state::{
-    Blake3ObjectIdentityV1, NativePrincipalContentStore, RuntimePrincipalStore, StateOwner,
+    Blake3ObjectIdentityV1, NativeContentStagingArea, NativePrincipalContentStore,
+    ReadyStagedContent, RuntimePrincipalStore, StagedContentId, StagedContentWriter, StateOwner,
     StateOwnerCodecV1, StateOwnerResolver, open_runtime_kv, open_runtime_principal_store,
 };
 
--- a/crates/astrid-storage/src/principal_state.rs
+++ b/crates/astrid-storage/src/principal_state.rs
@@ -7,9 +7,7 @@
 //! source.
 
 use std::cmp::Ordering;
-use std::fs::{File, OpenOptions};
-use std::io::Write;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use std::sync::Arc;
 
 use astrid_core::dirs::AstridHome;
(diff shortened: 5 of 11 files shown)


joshuajbouw:
Self-review of the two staging commits (`04a1b9a0`, `8855a254`). The shape is right: seal is the acknowledgement boundary with content-then-intent-then-rename durability ordering, sequence allocated at close so a slow old handle cannot shadow a newer one, staged arena frames unflushed until the next root commit's arena sync covers them (verified `persist()` is unchanged and its arena `sync_data` covers the whole prefix), and closure validation stays inside the root-CAS critical section so interrupted staging fails secure. The intent codec is versioned, domain-separated, and rejects trailing bytes; symlink/reparse and non-canonical names fail closed; unsealed entries quarantine instead of deleting. Good.

One crash-window bug that needs fixing before this merges, one lock-scope issue, and two notes.

**1. A crash inside `cleanup()` bricks the entire ready queue.** `remove_known_stage_files` removes `content.bin` first and the published marker last (the right order — the marker outlives the evidence), but `ready()` doesn't exploit it: it calls `load_ready` *before* `is_published`. Crash after `content.bin` is unlinked but before `remove_dir` completes, and the next scan hits a directory whose intent validates but whose content file is missing → `load_ready` errors → `ready()` propagates → every future `ready()` and, via the ordering guard, every future `publish()` fails. A pure crash-timing event, no attacker required, wedges all publication until manual surgery. The `durable_publication_marker_reaps_an_interrupted_cleanup` test covers only the marker-written-cleanup-not-started state where all three files still exist; the mid-cleanup states (intent+marker, marker-only) are the ones that brick.

Fix is a reorder: in `ready()`, read the published marker by path first; a valid marker means the entry is provably published regardless of completeness — finish the cleanup (the existing NotFound tolerance already handles partial removal) and skip `load_ready` entirely. Only unpublished entries go through full validation. Please add the two mid-cleanup states to the test.

**2. `stage_objects` hashes under the global engine mutex.** `stage_object` computes identity before taking the lock; `stage_objects` takes the lock first and then BLAKE3-hashes up to a 4 MiB batch inside the loop. During bulk ingest that serializes every KV commit on this engine behind content hashing. Compute the identities and encodings before acquiring the lock; only the index probes and the coalesced append need it.

**3. Note — deliberate whole-queue fail-closed on a corrupt entry.** A corrupt intent in `ready/` blocks the entire queue, not just that entry (`corrupt_ready_intent_fails_closed_without_deleting_staged_bytes` pins this). For acknowledged-but-unverifiable bytes I agree fail-closed is right for v1, but it means one bad sector in one intent halts all staged publication for every principal. When the writable provider ships this wants per-entry isolation with same-(owner,name) successors blocked and unrelated names flowing — noting it now so it's a decision, not an accident.

**4. Note — staging failure is a new road to engine poisoning.** An append failure during background ingest poisons the engine and takes all principal KV down with it until process restart. That's consistent with existing append semantics, but bulk ingest makes ENOSPC dramatically more reachable than a KV write ever did. This raises the priority of #1387 (in-process reopen after RequiresRecovery) to "before the writable provider ships".

Template check is red — there's a dangling empty bullet at the end of the Changes list.

With the `ready()` reorder and the lock-scope fix, this is mergeable after #1390 in stack order.

PR #1404, opened 2026-07-28
feat(storage): conserve deterministic computation in format v1

### LINKED ISSUE #1403: Add v1 conservation-of-computation substrate
## Summary

Add the format-v1 substrate for conserving deterministic computation over
content-addressed Astrid objects. The kernel must be able to identify a
semantics-visible invocation, verify its result, reuse durable evidence through
a disposable fleet memo index, and audit destructive storage maintenance
without making reasoning an authorization gate.

## Motivation

Astrid already gives stored values stable object identities. Deterministic
computations over those values should not be repeated when their exact
transform, inputs, parameters, and runtime semantics are unchanged. Reuse must
remain sound across engine releases, must not preserve external side effects,
and must not expose cross-principal memo or dedup state.

The same evidence machinery is also needed for background store maintenance and
for machine-recheckable garbage-collection decisions.

## Proposed Implementation

- Extend the existing format-v1 object grammar with canonical runtime profiles,
  invocation identities, derivation evidence, and linked GC plan/commit
  evidence.
- Bind invocation identity to the transform ObjectId, contract, ordered inputs,
  canonical parameters, semantics-visible runtime/ABI profile, execution class,
  and required snapshot or seed.
- Add Muninn as a bounded, policy-injected, disposable index over durable,
  engine-verified derivation evidence.
- Add a restricted Refinery pass boundary for integrity, attestation, sketching,
  and future compaction work without giving observers mutation authority.
- Require GC commit evidence to reference the exact proved plan and reject a
  changed fence-held snapshot.
- Document Forge, Huginn, Realm/distro, audit anchoring, sync, export, and
  public-content consumers of the same primitive.

This remains an amendment to format v1. It does not add a capsule-facing WIT
surface or introduce a second storage architecture.

## Acceptance Criteria

- Canonical encodings reject ambiguity and independently round-trip through the
  format reader.
- Memo conflicts become suspect evidence and never silently overwrite.
- Cache exhaustion or eviction cannot make a valid uncached computation fail.
- Refinery observers cannot implement the privileged compaction authority.
- A GC receipt cannot commit against a snapshot different from its proved plan.
- Workspace tests, all-features Clippy, formatting, and format-v1 reader checks
  pass.

joshuajbouw:
## Linked Issue

Closes #1403

## Summary

Adds Astrid's format-v1 substrate for conserving deterministic computation over
content-addressed objects. An invocation now has a canonical identity over the
exact transform, inputs, parameters, semantics-visible runtime profile, and
execution context; verified results can be reused through Muninn without making
the cache authoritative.

This also adds the restricted Refinery pass boundary and machine-recheckable GC
plan/commit receipts. All additions extend the existing v1 format. There is no
new capsule-facing WIT surface.

## Changes

- add canonical runtime-profile and invocation records, including explicit
  execution classes that prevent memoizing external side effects
- add durable derivation evidence and a bounded, sharing-domain-partitioned
  Muninn index whose eviction only affects performance
- add a Refinery observer interface over prevalidated immutable objects, with
  compaction authority sealed inside the engine
- add linked GC plan and commit evidence with snapshot-digest equality enforced
  at commit construction
- amend the format-v1 Rosetta specification and independent reader for the new
  object kinds while preserving crash-safe upgrade from the earlier v1 digest
- document conservation of computation, Muninn, Huginn, Refinery, freeze-audit,
  and implementation triage
- connect Forge, Realm images, distros, audit anchoring, synchronization,
  compliance export, and future public-content representations to the same
  derivation primitive
- add regression tests for canonical identity, memo conflict handling,
  cache-independence, Refinery authority separation, and GC receipt binding
- make Muninn closure validation iterative, preserve conflict guards across
  eviction/rebuild, and require evidence for trust rehabilitation
- decode and reject every new v1 schema independently using Rust-produced
  valid fixtures and identity-consistent malformed fixtures
- isolate Rosetta amendment handling in a dedicated module, keeping the native
  principal-state integration below the repository source-size cap

## Verification

- `cargo test --workspace -- --quiet`
- `cargo clippy --workspace --all-features -- -D warnings`
- `git diff --check origin/main...HEAD`
- format-v1 Rosetta and independent Python reader tests
- targeted storage-model, storage-engine, Muninn, Refinery, and GC evidence
  suites
- all eight commits verified with `git verify-commit`

## Checklist

- [x] Linked to an issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,10 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
   Refinery contract and initial engine seam unify scrub, SHA-384 cross-hashing,
   resemblance sketches, and representation-aware compaction behind one
   resource-budget shape while retaining a sealed engine-only mutation
-  boundary. Durable storage freeze and
+  boundary. Linked GC plan and commit evidence keep the exact fact snapshot,
+  retention policy, auditor proof, condemned set, and placement transition
+  replayable, while commit construction rejects a changed fence-held snapshot.
+  Durable storage freeze and
   freedom-to-operate records pin the associated format, chunking, identity,
   and public-layer research decisions without activating a capsule or WIT
   surface.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,10 +21,11 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
   specified as a disposable fleet index over durable derivation evidence, with
   differential spot checks, logical-versus-physical compute accounting, and
   principal/trust-domain authority preserved below its cache. Huginn applies
-  the same primitive to canonical local-model context and prefix assembly. One
-  Refinery scheduler and resource budget cover scrub, SHA-384 cross-hashing,
-  resemblance sketches, and representation-aware compaction while retaining a
-  sealed engine-only mutation boundary. Durable storage freeze and
+  the same primitive to canonical local-model context and prefix assembly. The
+  Refinery contract and initial engine seam unify scrub, SHA-384 cross-hashing,
+  resemblance sketches, and representation-aware compaction behind one
+  resource-budget shape while retaining a sealed engine-only mutation
+  boundary. Durable storage freeze and
   freedom-to-operate records pin the associated format, chunking, identity,
   and public-layer research decisions without activating a capsule or WIT
   surface.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,10 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
 
 - **Astrid has a content-addressed computation design contract.** A canonical
   invocation object now specifies the complete identity boundary for reusable
-  deterministic work, including typed ordered inputs, canonical parameters,
-  semantics-visible runtime profiles, output contracts, provenance snapshots,
-  and explicit seeds. Pure, snapshot-bound, effectful, and nondeterministic
+  deterministic work, including the exact transform closure, its registered
+  contract, typed ordered inputs, canonical parameters, semantics-visible
+  runtime profiles, output contracts, provenance snapshots, and explicit
+  seeds. Pure, snapshot-bound, effectful, and nondeterministic
   execution classes prevent memoization from replaying side effects. Muninn is
   specified as a disposable fleet index over durable derivation evidence, with
   differential spot checks, logical-versus-physical compute accounting, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,22 @@ Changelog tracking starts with 0.2.0. Prior versions were not tracked.
 
 ### Added
 
+- **Astrid has a content-addressed computation design contract.** A canonical
+  invocation object now specifies the complete identity boundary for reusable
+  deterministic work, including typed ordered inputs, canonical parameters,
+  semantics-visible runtime profiles, output contracts, provenance snapshots,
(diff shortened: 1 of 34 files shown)


joshuajbouw:
Self-review at maximum depth — this is the most consequence-dense artifact of the program, so every canonical path got the adversarial pass.

**What holds, verified against the code:**

- **Second-representation attacks are structurally dead.** Every `from_object_record` re-encodes and requires byte-exact equality with the input record (`derivation.rs:462,761`; same pattern in the plan decode). Combined with strict-ordering requirements on proposals/host functions and ordinal-checked input labels, there is exactly one accepted encoding per logical value. This is the single most important property of the invocation contract and it is enforced in the right place.
- **The runtime profile pins the semantics-visible surface as content-addressed contracts** — including `resource_failure_semantics`, which is a subtle and correct catch (fuel/OOM behaviour is semantics-visible). Engine builds sharing the surface share the profile id; memos survive engine upgrades exactly as intended.
- **Execution-class invariants enforced twice** (construction and decode): pure+snapshot rejected, snapshot-bound requires snapshot, only Pure/SnapshotBound memoizable — checked again independently in `verify_derivation_evidence`, so a malformed class can't reach the index even through a forged record.
- **Muninn is constitutionally disposable and verify-before-admit**: every identity recomputed, the complete output Owns-closure walked with cycle detection and extraneous-record rejection, conflicts mark the existing entry Suspect and never replace (with the incoming never indexed — both effectively quarantined), lookups serve only Verified, capacity injected with no library default.
- **GC receipts implement the binding exactly as required**: `GcCommitEvidence::new` takes the fence-recomputed snapshot digest and fails closed on mismatch with the plan (`FenceSnapshotChanged`); the receipt carries the plan's ObjectId; `validate_plan` gives auditors the recheck. No-op commits rejected.
- **Refinery privilege is typed, not configured**: `EngineCompactionPass` behind a sealed supertrait — unimplementable outside the engine — while observer passes can only propose Evidence/Derived records through the sink. Exactly the "one scheduler, two authority levels" split.

**One fix required before merge:**

1. **`visit_output` recursion is attacker-depth-bounded** (`muninn.rs:372`). The output closure shape is produced by transform capsules — attacker-influenceable — and a deep Owns chain (a 100k-link list is cheap to emit) overflows the verifier's stack: a crash in the engine, triggered by data, during admission. Convert the closure walk to an explicit work-list (the marks map already gives you everything needed) or enforce a declared depth bound from `RecoveryLimits`. Same class as the bounded-traversal fix we already made in content DAG reads (afc6cf08) — this is that bug's sibling.

**Two sharpenings, small:**

2. `set_trust_state` currently permits Suspect→Verified with `spot_check_evidence: None`. Rehabilitation should require the audit record: make upgrades *to* Verified demand a `Some` evidence id (downgrades may stay evidence-optional — you don't need paperwork to stop trusting). One arm of a match.

3. Two documentation notes worth a sentence each where they belong: (a) invocation inputs are Evidence-kind references (correct — memos shouldn't pin input bytes into an owner's closure), with the consequence that spot-checks are only executable while the input closure remains live — the refinery's sampling scheduler should prefer young entries; (b) runtime-profile and semantic-contract *registration* is signature-level authority per the semantic-representations rules — a profile whose host functions are secretly nondeterministic is a registration failure, not a model-checkable property, and the doc should say which authority owns that risk.

Also noting for the follow-up queue, not this PR: the in-memory index at capacity is first-come-kept-forever (no eviction policy until the resource authority lands) — fine for v1, worth a comment so nobody mistakes it for LRU.

With the traversal fix and the trust-upgrade guard, this is merge-worthy, and the contract is worthy of what's going to be built on it.

PR #1405, opened 2026-07-28
fix(capsule): discard trap-poisoned pooled instances + dispatch-failed signal + per-principal watchdog (2026-07-28 wedge)

vrogojin:
## Summary

**SEV robustness fix** for the 2026-07-28 all-principals agent wedge on a concierge dev deployment (base `c511f5f`): one epoch-interrupted runaway LLM stream permanently wedged every agent turn for every principal until a daemon restart. Three defense layers, one commit each:

### 1. `fix(capsule)`: discard trap-poisoned pooled instances (the wedge itself)

Wasmtime **poisons a component instance on any non-clean guest exit** (trap, panic-unwind, future-drop mid-call): every later call fails `cannot enter component instance`. `PoolCheckout::drop` returned such instances to the pool unconditionally, and **LIFO checkout re-leases the poisoned instance first** — it fails instantly, is returned again, and stays the hottest slot. One trapped invocation therefore permanently captured the capsule's whole invocation path (observed: a 300s epoch-interrupted `openai-compat` stream at 11:19:42Z → every later `handle_llm_request`/`llm_describe` failing `cannot enter component instance` → all turns for all principals silently dead until restart).

- `PoolCheckout` gains an **armed** flag: `arm()` immediately before guest entry, `disarm()` only on a clean exit (`Ok`, or export-lookup failure where the guest was never entered). `Drop` of an armed checkout **discards** the instance; the released permit lets a later checkout lazily rebuild. Covers trap, panic, and cancellation paths alike.
- `checkout()` now rebuilds on an empty warm set for **every** pool, including the size-1 `host_process` carve-out: holding the permit guarantees total ≤ max (the exact invariant the old fail-closed branch protected), and refusing left a discarded carve-out permanently un-invocable.
- Regression tests: `armed_checkout_is_discarded_and_pool_rebuilds`, `disarmed_checkout_returns_to_pool`, `carveout_pool_rebuilds_after_discard`.

### 2. `feat(capsule)`: publish `astrid.v1.dispatch.failed` on invocation errors

Dispatch is fire-and-forget; a failed invocation was only a host-side `warn!`, so the origin (react awaiting the `llm.v1.request.generate.*` it published) waited forever. Both dispatch paths now publish `astrid.v1.dispatch.failed` `{ failed_topic, capsule_id, action, error, request_id?, session_id? }` (correlation ids best-effort-extracted from the failed event's payload, parse capped at 512 KB), stamped with the failed event's principal so a per-principal subscriber handles it under the right KV scope. Loop guard: failures of this topic's own handling are never re-signalled. Tests: emission + principal + correlation, loop guard, extractor shapes.

### 3. `fix(kernel)`: per-principal react watchdog ticks

`astrid.v1.watchdog.tick` was published principal-less, so react's `handle_watchdog_tick` only ever saw the capsule **owner's** `react.active_sessions` — per-principal sessions were invisible to phase-timeout recovery (the backstop that should have failed the wedged turns in 30–60 s). The watchdog task now watches `user.v1.prompt` and re-publishes each 5 s tick once per principal seen within a 15-min TTL (bounded set, cap 256, stalest evicted), keeping the principal-less tick for owner-scope sessions. Test: eviction/bounding discipline.

## Validation

- `cargo test -p astrid-capsule --lib` → **524 passed, 0 failed** (incl. 6 new); `cargo test -p astrid-kernel --lib` → **202 passed** (incl. 1 new); `cargo clippy -p astrid-capsule -p astrid-kernel --lib --tests` → clean (workspace denies incl. `arithmetic_side_effects`).
- Deployed to the concierge dev stack (as kernel patches on `c511f5f`) and validated live: full agent turns stream end-to-end (ready → deltas → tool round-trip → final reply, transcript + token usage persisted).

## Incident reference

Full RCA (timeline, daemon-log evidence, wasmtime semantics) in the concierge deployment's incident notes, 2026-07-28: runaway `!` degeneration → 300 s epoch interrupt → poisoned instance → LIFO capture → silent global hang. This PR removes the wedge class (1), makes any future capsule breakage fail loudly in ~1 s (2), and restores the *generic* turn-recovery backstop (3).

**HELD — do not merge without maintainer review.** Opened from a fork (`vrogojin/astrid`); author has read-only access to this repo.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/astrid-capsule/src/dispatcher.rs
+++ b/crates/astrid-capsule/src/dispatcher.rs
@@ -406,6 +406,7 @@ impl EventDispatcher {
             dispatch_to_capsule_queues(
                 &capsule_queues,
                 &self.chain_locks,
+                &self.event_bus,
                 matches,
                 topic,
                 payload_bytes,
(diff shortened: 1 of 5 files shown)


joshuajbouw:
Please make sure that you are:
a) signing off on DCO, your agent can't and shouldn't do this for you
b) rebasing
c) following the PR template

I see that Claude is being used as being the one making the commits on its own, but it is critical for PR owners to understand the changes that are being made in full.

PR #1381, opened 2026-07-26
fix(capsule): make component→root capability merge exhaustive (#1232)

jvsteiner:
## Linked Issue

Closes #1232.

## Summary

`discovery.rs` folded each `[[component]].capabilities` into the root manifest field-by-field by hand-enumeration, so fields not explicitly listed were silently dropped. The security gate reads only the root, meaning a component-only declaration could appear in `astrid capsule show` while granting nothing.

## Changes

- Add exhaustive `CapabilitiesDef::merge_from`: allowlists concatenate and flags are OR-ed.
- Destructure the source without `..`, so adding a capability field fails compilation until its merge behavior is handled.
- Replace the partial discovery merge with the exhaustive helper.
- Add discovery-seam and complete flag/list regressions, including component-only positive grants and preservation of existing root grants.

## Verification

- `cargo fmt --all -- --check`
- `cargo test -p astrid-capsule-types merge_from_ -- --nocapture`
- `cargo test -p astrid-capsule component_level_capabilities_merge_into_root -- --nocapture`
- `cargo clippy -p astrid-capsule-types -p astrid-capsule --all-features -- -D warnings`
- `git diff --cached --check`

## AI / Tool Assistance

Assisted-by: Anthropic Claude: Opus 4.8
Assisted-by: OpenAI Codex: GPT-5

Claude assisted Jamie with the original implementation. Codex independently reviewed the architecture and capability semantics, strengthened the regression coverage, reconciled the patch onto current `main`, and validated the affected crates. I reviewed the final combined patch and created the GPG-signed, DCO-compliant commit. Jamie and Claude remain credited as co-authors.

## Checklist

- [x] Linked to an issue
- [x] CHANGELOG.md updated
- [x] I understand every change in this PR and can explain its design, risks, and validation.
- [x] I reviewed and tested any meaningful tool-generated output included in this PR.
- [x] Every non-bot, non-merge commit has a matching `Signed-off-by` trailer.

PR #1321, opened 2026-07-23
feat(capsule): inbound TCP bind (bind_tcp) + per-principal run-loop resources + concurrent workers

jvsteiner:
Capsule **inbound TCP bind** (`bind_tcp`) + **per-principal resources** for run-loop capsules + **concurrent run-loop workers** for TCP-server capsules.

### What & why
Enables a capsule to bind an inbound TCP port and serve it from its run loop — the runtime primitive the `srouter` capsule needs (Anthropic-Messages loopback ingress on `127.0.0.1:8788`). Three commits:

1. `feat(net): implement capsule inbound TCP bind (bind_tcp)` — the host-side `bind_tcp` + manifest `net_bind` capability + `manifest_gate` enforcement.
2. `feat(capsule): deliver per-principal resources to run-loop capsules` — run-loop capsules receive the owner principal's resources (overlaps the run-loop-KV gap tracked in #1197).
3. `feat(capsule): concurrent run-loop workers for TCP server capsules` — a run-loop capsule can serve concurrent connections.

### Rebase note
Originally authored on old main `6ea3489`; **cleanly cherry-picked onto current `main` (41 commits later)** with zero conflicts — git auto-followed the `capabilities.rs` relocation (`astrid-capsule` → `astrid-capsule-types`).

### Verification
- `cargo check -p astrid-capsule` clean.
- `cargo test -p astrid-capsule --lib` → **578 passed, 0 failed**.

Draft pending author review of the replay onto the reorganized tree.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1540, opened 2026-08-15
feat(capsule): concurrent run-loop workers for TCP server capsules (#1231)

### LINKED ISSUE #1231: feat(capsule): concurrent workers for run-loop TCP-server capsules (single-Store run-loop serializes requests)
## Summary

A run-loop (`#[astrid::run]`) capsule that hosts a loopback TCP server (via `bind_tcp`) is pinned to a **single Store**, so it handles inbound connections **serially** — while it blocks on one request's upstream I/O, every other connection waits. A client that fans out parallel requests (e.g. Claude Code with subagents, pointed at a capsule-hosted Anthropic-shim gateway) queues behind each other. Measured: 5 concurrent requests completed at 2/4/6/8/10 s (one-at-a-time) instead of ~2 s each.

The wasm guest is single-threaded with blocking host I/O (`http::send`, `HttpStream::read_chunk` block the one thread), so cooperative concurrency inside one instance is impossible. Concurrency requires **N worker Stores**.

## Why not `SO_REUSEPORT` (important platform finding)

The obvious "N instances each bind the same port with `SO_REUSEPORT`" does **not** work on macOS: Darwin permits multiple `SO_REUSEPORT` binds but delivers **every** new connection to the *most-recently-bound* socket (no `SO_REUSEPORT_LB`; that's FreeBSD 12+). So N workers collapse to 1 effective worker on macOS — it silently "works" (all binds succeed) but adds zero concurrency. Linux (≥3.9) hash-distributes, so it's fine there, but a macOS dev/host is common.

## Recommended architecture (Approach B — shared listener, cross-platform)

Bind the loopback port **once**; share the `Arc<tokio::net::TcpListener>` across N worker Stores; each worker runs `run()` and blocks on `accept()` on the **same** socket. `tokio::net::TcpListener::accept` is `&self`-safe under concurrent callers — the kernel hands each `accept()` exactly one connection, so the OS accept queue is the load balancer. Identical behavior on macOS and Linux, and it reuses the existing `Arc<TcpListener>` slot verbatim.

## Reference implementation (built + verified)

Implemented on our `bind_tcp` branch; **5 concurrent requests now complete in ~2.4–3.1 s each (parallel); 8 workers spawn; clean teardown across restarts; N=1 is byte-identical to today.** Files (paths under `crates/`):

1. **`astrid-capsule-types/src/manifest/capabilities.rs`** — add to `CapabilitiesDef`:
   ```rust
   /// Concurrent run-loop workers for a loopback TCP server capsule. Each is an
   /// independent Store running `run()` and accepting on the shared bound
   /// listener. None/1 = today's single-instance behavior. Only honored for
   /// run-loop capsules that declare net_bind and no host_process.
   #[serde(default)]
   pub bind_workers: Option<usize>,
   ```

2. **`astrid-capsule/src/discovery.rs`** (the component→root capability merge, ~line 235-246) — **must** promote the new field, or it silently arrives `None` (see the companion "capability-merge footgun" issue):
   ```rust
   if caps.bind_workers.is_some() { manifest.capabilities.bind_workers = caps.bind_workers; }
   ```

3. **`astrid-capsule/src/engine/wasm/host_state.rs`** — add `shared_listeners: Arc<dashmap::DashMap<(String,u16), Arc<tokio::net::TcpListener>>>` beside `connection_principals`; clone it into every HostState in `make_state` (dashmap is already a dep).

4. **`astrid-capsule/src/engine/wasm/host/net/mod.rs`** (`bind_tcp`) — after the capability gate + loopback rail, resolve the listener via the shared map: first worker binds (under the shard lock, which serializes racing workers — needed because macOS has no `SO_REUSEADDR` fast-rebind), siblings clone the `Arc`. `TcpListenerSlot` already stores `Arc<TcpListener>`; accept/poll_accept unchanged.

5. **`astrid-capsule/src/engine/wasm/mod.rs`** (the run-loop load path):
   - Resolve `worker_count = if has_run_export && host_process.is_empty() && !net_bind.is_empty() { bind_workers.unwrap_or(1).clamp(1, runtime_limits.instance_pool_size) } else { 1 }`. If `worker_count > 1` **and** the capsule declares interceptors, force 1 with a warn (N subscriptions would double-process events).
   - Build `worker_count` warm instances (was `pool_min_idle`); factor the fuel/epoch setup into a per-store helper; collect `Vec<(Arc<AsyncMutex<Store>>, Instance)>`.
   - Fold the three single-store install steps (`ready_tx`, interceptor auto-subscribe, and the owner-context overlay from the #1224 fix) into **one per-worker loop** so each Store is fully provisioned before its run task spawns.
   - `run_handle: Option<JoinHandle>` → `run_handles: Vec<JoinHandle>`; `ready_rx: Option<Mutex<Receiver>>` → `ready_rxs: Vec<...>`. Spawn one run task per worker; the single shared `cancel_token` (already cloned into each HostState) fans out.
   - `wait_ready` awaits **all** N receivers within one timeout budget. `unload` keeps the single `cancel_token.cancel()` then `for h in run_handles.drain(..) { h.abort() }`.

**No regression by construction:** `bind_workers` defaults to `None` → `worker_count = 1` → one warm instance, one run task, 1-element Vecs, shared map with one entry never reused → identical to the current path. `host_process` run-loops and non-`net_bind` daemons are hard-pinned to 1 so their side-effect-bearing singletons never replicate.

## Caveat to document

Replicating `run()` N times duplicates all of its side effects (IPC publishes, singletons, interceptor deliveries). Worker scaling is therefore opt-in and gated to **stateless connection-handler** run-loops (declares `net_bind`, no interceptors) — never applied to arbitrary daemons.

Happy to open a PR.

jvsteiner:
## Linked Issue

Closes #1231

## Summary

A run-loop (`#[astrid::run]`) capsule hosting a loopback TCP server is pinned to a **single Store**, so it serves connections serially — while it blocks on one request's upstream I/O, every other connection waits. The guest is single-threaded with blocking host I/O, so cooperative concurrency inside one instance is impossible; the only route is N instances.

`bind_workers` lets such a capsule run N worker Stores, each executing `run()` against **one** shared bound listener and blocking on `accept()` — the OS accept queue load-balances. `N = 1` is byte-identical to current behaviour.

Lands on top of #1380, as set out when #1321 was split into focused PRs. Rebased onto current `main` rather than cherry-picked: the run-loop lifecycle was rebuilt underneath the original branch (activation watch channel, mid-run cancellation racing `call_async`, run loop as an async task), so this is a reimplementation against that machinery.

## Changes

- **`bind_workers` (`[capabilities]`)** — a run-loop capsule declaring `net_bind` and no `host_process` runs N worker Stores, clamped to `instance_pool_size`. Forced to 1 with a warning when the capsule declares interceptors, since N subscriptions would double-process every event.
- **Shared listener (Approach B)** — the first worker to `bind_tcp` binds and inserts into a per-capsule `DashMap<(host, port), Arc<TcpListener>>`; siblings clone the `Arc`. The bind runs under the shard lock so racing workers serialize rather than failing `EADDRINUSE` (macOS sets no `SO_REUSEADDR`). `SO_REUSEPORT` was rejected: on macOS it does not load-balance, it delivers every connection to the most-recent bind.
- **Per-worker activation** — one `watch::Sender`; each worker takes its own `subscribe()` receiver, so a single publish releases all N. A worker cancelled before activation returns without ever locking its Store.
- **Per-worker cancellation and teardown** — the same `cancel_token` clone to every worker, each keeping the existing `biased` `select!` so a compute-bound guest unwinds. `run_handles`/`ready_rxs` become `Vec`s; `unload` aborts all, `wait_ready` awaits all.
- **Owner context on every worker Store** — #1380's `install_run_loop_owner_context` is applied to each, not just the first; a worker without it would serve with no principal authority.
- **A wildcard port is never shared** — port `0` means "any ephemeral port", so two such requests are different addresses. Keying the registry on the requested port conflated them, and a pooled capsule binding port 0 four times would have received one socket four times. Caught by the existing quota test.
- **The listener quota bounds sockets, not workers** — `MAX_ACTIVE_TCP_LISTENERS` is 4, so charging per worker would let `bind_workers = 8` exhaust it with a single port. The charge is taken only when a socket is created, and `TcpListenerSlot::listener_count` became `Option<_>` so only the binding worker releases it.
- **`localhost` is normalized before the registry key** — otherwise two workers naming the same address produce two entries and race for one OS port.
- **`discovery.rs` loses its hand-promotion of `bind_workers`** — #1381 replaced the field-by-field merge with an exhaustive destructure, so `merge_from` gains one arm and the workaround is deleted.
- **`bind_workers` is deliberately not a held capability** — it grants no authority, it parameterises an already-granted `net_bind`. Listing a worker count in `held_names` would misreport the security posture in `astrid capsule show` and the audit trail, so `held_names_and_has_agree_when_all_held` excludes it by name.
- **New fixture and harness** — `e2e/fixtures/astrid-capsule-concurrency` and `e2e/concurrency.sh`.

## Verification

`cargo test -p astrid-capsule -p astrid-capsule-types` — 688 tests, all green. `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean.

**End-to-end.** `e2e/concurrency.sh` runs the fixture (which blocks 500 ms per request) twice against an isolated `ASTRID_HOME`, changing one manifest line between runs:

```
astrid#1231: serial baseline (bind_workers = 1)
  elapsed ms: 504 1007 1510 2013 2515
astrid#1231: concurrent (bind_workers = 5)
  elapsed ms: 505 505 505 505 505
astrid#1231: asserting
  baseline spread   2011ms  (>= 500ms, serialized)
  concurrent spread 0ms  (<  500ms, parallel)
astrid#1231: asserting every worker died
  connect refused on 18231 — no worker survived the stop
astrid#1231: PASS
```

The baseline is the issue's own symptom reproduced: each client queues ~505 ms behind the one before it.

To run it:

```sh
cargo build --bin astrid --bin astrid-daemon
(cd e2e/fixtures/astrid-capsule-concurrency && cargo build --release)
./e2e/concurrency.sh
```

**Not wired into CI** — it needs a built runtime, a wasm fixture and a live daemon, and the existing `wasm_e2e` fixtures are prebuilt artifacts absent from the tree. Happy to wire it up if you want that; it seemed like its own change.

Two deliberate choices in the harness, both of which bit me first:

- It asserts **completion spread**, not absolute wall-clock. A shared runner can be slow without being serial, and "all five finished within one request-time of each other" is the property under test.
- Teardown is asserted by **connecting, not binding**. A bind probe is confounded by `TIME_WAIT` from the client connections and reported a surviving worker when `lsof` showed nothing listening at all.

**Unit coverage:**

| Claim | Test |
|---|---|
| Workers share one bound socket | `worker_stores_dedupe_onto_one_bound_socket` |
| `localhost` and the literal are one address | `localhost_and_loopback_literal_share_one_registry_entry` |
| Quota charged once, released once | the two above, plus the existing quota test |
| A wildcard port never shares | `tcp_listener_quota_is_independent_and_released_on_drop` (existing) |
| `bind_workers` merges component→root | `merge_from_unions_every_field` |
| A silent component does not erase it | `merge_from_preserves_root_bind_workers_when_component_is_silent` |

## AI / Tool Assistance

Assisted-by: CLAUDE:claude-opus-4.8

Substantial. The rebase-and-reimplementation, the fixture capsule, the harness, the tests and this description were all produced with Claude, working from a stale branch (`feat/capsule-bind-tcp-on-main`, authored against old `main`) that no longer applied.

How it was reviewed and validated:

- **Every claim in Verification was executed, not asserted.** The numbers above are real output from this branch on an arm64 macOS host, not reconstructed.
- **Three defects were found by tests rather than by reading**, and each is called out in Changes above: the wildcard-port sharing bug (caught by the *existing* quota test), the per-worker quota charge, and the `localhost` registry key. The first would have shipped.
- **Two mistakes were made and corrected during the work.** A scripted conflict resolution silently deleted #1380's `install_run_loop_owner_context`; caught by checking the symbol still existed, then the whole file was redone bottom-up. And the first teardown assertion was a false positive from `TIME_WAIT`, which would have sent a reviewer hunting a bug that does not exist.
- **The three exhaustiveness guards in this codebase did real work** — `merge_from`'s no-`..` destructure, `merge_from_unions_every_field`'s fully-populated literals, and `held_names_and_has_agree_when_all_held`. Each forced a decision rather than letting the change slide through, and the `held_names` one is recorded in the code where the next person will hit it.

I can explain every hunk, and the two merge decisions I would most want scrutinised are the per-worker owner context and the per-worker activation subscribe.

## Checklist

- [x] Linked to an issue


joshuajbouw:
Good catch and related to what I had closed recently. Sometimes you find the best issues when actually building things with the project. Ok - will look into it.

joshuajbouw:
Note the signature on the commit is unverified.

PR #1543, opened 2026-08-16
Harden public runtime boundaries and resource accounting

joshuajbouw:
## Summary

Public-scope runtime hardening from the adversarial review pass:

- Bound adversarial capsule resource use: WASM table growth, concurrent in-flight CPU fuel, host-process file injection size, and entry-count/size caps where applicable.
- Pin approved capsule executables so an approved manifest cannot be repointed at different content-addressed bytes after authority approval.
- Honor the uplink keypair method gate: a retained device key cannot authenticate a principal after keypair authentication is removed from the profile's active methods.
- Recover storage group leadership after a leader panic instead of leaving the group wedged.
- Expose semantic capability grants so consent and introspection surfaces render action, scope, and impact instead of raw manifest field names; malformed capability metadata fails closed rather than rendering an empty permission list.
- Add regression coverage pinning principal-owned IPC subscriptions, workspace-relative approval boundaries, event topic depth/accessors, resource-pattern semantics, semantic expansion translations, and archive provenance invariants.

## Verification

- `cargo check` across `astrid-capsule`, `astrid-uplink`, `astrid-storage`, `astrid-workspace`, `astrid-events`, `astrid-capabilities`, and `astrid-mcp` (all targets) passes.
- Full workspace test/clippy/fmt verified on the parent hardening branch before carving; this branch is the public-scope subset.

## Linked Issue

No tracked issue - traceability via this PR, the per-fix changelog entries, and the signed commit history.


joshuajbouw:
Closing in favor of per-topic PRs against the audit sub-issues (#1545-#1552), each following the PR template and title conventions. The compilation shape and the DCO mismatch on the changelog commit are both addressed in the split.

PR #1819, opened 2026-09-01
fix(cli): isolate run idle client policy

### LINKED ISSUE #1812: fix(cli): make astrid run idle timeout configurable
## Problem

`astrid run` currently terminates after 120 seconds with no daemon message. The duration is hardcoded in `crates/astrid-cli/src/commands/headless.rs`, and `astrid run --help` exposes no override.

This is distinct from the outbound HTTP timeouts. `[http].header_deadline_secs` and `[http].stream_read_timeout_secs` govern capsule HTTP behavior; raising them cannot prevent the headless CLI from exiting after its own 120-second idle wait.

Real sandboxed tools can legitimately exceed 120 seconds during a cold upstream start. The operator and invoking user need an explicit, bounded way to opt into a longer wait without weakening other HTTP or daemon policy.

## Required outcome

- Add an operator default for the headless response idle timeout under `[timeouts]`, preserving 120 seconds when absent.
- Add an explicit per-run override to `astrid run` whose name describes idle/no-message semantics; `--idle-timeout-secs` is preferred, with `--timeout` accepted only as a documented alias if retained for ergonomics.
- Validate zero, overflow, and unreasonable values fail closed; do not silently convert the setting into an unbounded wait.
- Reset the idle timer when a valid message for the active run is received. Do not reinterpret it as a whole-request deadline.
- Return the established timeout exit status without calling `std::process::exit` inside the response collector, so behavior is unit-testable and cleanup/disconnect can run.
- Keep outbound HTTP controls independent. Document when to use per-request `first-byte-ms`, `between-bytes-ms`, and `total-ms` versus the `astrid run` idle timeout.
- Add CLI parse/config precedence/default tests and a delayed-response regression exceeding a short injected timeout.
- Add a changelog fragment for Astrid 0.11.0.

## Claim limits

This does not add a generic direct-tool CLI or a raw IPC firehose. It does not change capsule HTTP defaults, capability enforcement, approval behavior, public WIT, or daemon shutdown.

## Release gate

Must land on `main` and receive independent exact-head full GLM-5.3 Max plus Luna Max review before the Astrid 0.11.0 release cut.

### LINKED ISSUE #1816: fix(config): separate client, bootstrap, and runtime configuration authority
## Problem

Astrid currently has one broad configuration loader tied to `ASTRID_HOME`, while AOS 2026.9.0 requires stopped `~/.aos/runtime/` to contain exactly `astrid.volume`. Treating every setting as one host TOML either makes startup depend on mounting the volume or moves runtime/OS authority into an unauthenticated host-side file.

Issue #1812 exposed the missing distinction: `astrid run` response-idle timeout is a **client behavior**, not daemon, kernel, storage, or OS policy.

## Configuration planes

1. **Client configuration (pre-mount, non-authoritative):** CLI UX and connection behavior such as `run_idle_secs`, output preferences, endpoint selection, and local diagnostic behavior. Available before the volume is mounted. Under AOS this belongs outside runtime state, canonically `~/.aos/etc/astrid/client.toml` (or an equally explicit platform config path). Standalone hosted Astrid uses its platform user-config location. A per-command flag has highest precedence.
2. **Bootstrap configuration (pre-mount, narrowly authoritative):** release selection, volume location, mount/provider selection, and the minimum material needed to start and authenticate the runtime. This is AOS/host bootstrap state, not general Astrid runtime policy.
3. **Runtime/OS configuration (post-admission):** capability policy, principals, model/provider policy, budgets, groups, and other governed state. This remains inside authenticated Astrid storage or a signed System Generation on native Astrid OS; it must not be silently moved into a mutable host TOML.

## Required outcome

- Reclassify #1812 `run_idle_secs` as client configuration rather than general runtime policy.
- Add an explicit client-config discovery/input independent of `ASTRID_HOME` and `astrid.volume`; prefer an accurately named `ASTRID_CLIENT_CONFIG_PATH` if an environment override is required.
- AOS launches its Astrid CLI with the client-config path outside `~/.aos/runtime/`.
- Preserve CLI `--idle-timeout-secs` as the highest-precedence per-run override.
- Do not externalize the complete `astrid-config::Config` or runtime security policy.
- Reconcile the existing mismatch where loader discovery reads `$ASTRID_HOME/config.toml` while `astrid config edit` writes `$ASTRID_HOME/etc/config.toml`; migration must be bounded and non-ambiguous.
- Stopped `~/.aos/runtime/` remains exactly `astrid.volume`, and client help/config works while the volume is absent or unmounted.

## Native OS mapping

On standalone Astrid OS, the same semantic setting is a per-user shell/client preference once the OS configuration service is available. Boot-critical policy belongs to the signed System Generation/bootstrap path; runtime policy belongs to authenticated OS state. A host filesystem path is an adapter detail, not the native contract.

## Falsifiers

- `astrid run` reads its configured idle timeout with `astrid.volume` absent and the daemon stopped.
- Changing client timeout cannot alter daemon HTTP policy, capabilities, principals, budgets, or runtime admission.
- Runtime policy cannot be injected through `ASTRID_CLIENT_CONFIG_PATH`.
- Conflicting legacy/canonical files cannot silently select different authorities.
- Relative, symlinked, foreign-owned, or overly permissive explicit client paths fail closed.
- CLI and daemon do not accidentally interpret the same field as different authority classes.

## Release relationship

Main-targeted architecture/correctness companion for #1812 and the Astrid 0.11.0 / AOS 2026.9.0 release journey. PR #1815 is not release-ready until its persistent configuration is classified correctly. No release, tag, publication, or broad config migration authority is implied.

joshuajbouw:
## Linked Issues

Closes #1812 and closes #1816.

## Summary

- Moves `run_idle_secs` out of the broad runtime `Config` into a narrow pre-mount client configuration plane.
- Supports explicit `ASTRID_CLIENT_CONFIG_PATH` and the canonical `~/.aos/etc/astrid/client.toml` path, independent of `ASTRID_HOME`.
- Rejects relative, redirected, foreign-owned, permissive, oversized, or malformed explicit client files; unknown runtime policy keys fail closed.
- Preserves the 120-second default, zero/above-one-day rejection, CLI-over-client-config precedence, and exit 53 at the CLI boundary.
- Session-bearing `agent.v1.stream.delta` frames now reset the idle deadline. Tool, stream, and approval control frames must be correlated to the run; foreign same-principal approvals neither reset nor receive responses.
- Adds bounded best-effort disconnect and approval writes so a wedged send cannot defer timeout determinism.
- Splits headless tests into a separate module and uses deterministic source-boundary timeout/message injection without scheduler-sensitive sleeps.

## Test Plan

- New client-path, private-file, symlink, permissive-mode, malformed-runtime-key, precedence, and timeout-bound tests.
- New stream-delta, foreign-frame, foreign same-principal approval, run-correlated approval, timeout cleanup, and wedged-send tests.
- Focused regression coverage for parser precedence and client/default resolution.

## Verification

- `cargo fmt --check`
- `cargo test -p astrid-config --lib` — 126 passed
- `cargo test -p astrid --bin astrid` — 642 passed, 2 ignored (release-only)
- `cargo clippy -p astrid-config --all-targets -- -D warnings`
- `cargo clippy -p astrid --all-targets -- -D warnings`

## Residuals

- Current host approval producers emit uncorrelated/nil source IDs, so this fail-closed CLI consumer cannot safely answer them. Restoring same-principal auto-answer safely requires session/run ownership to propagate at routing/types boundaries; that propagation is outside this exclusive file set.
- CI is not terminal from this local seat.
- AOS launcher export of the canonical client path remains a later adapter slice.

## Checklist

- [x] Linked to an issue
- [x] Changelog fragment updated under the allowed path
- [x] Every commit is GPG-signed with a matching DCO trailer
- [x] No merge, tag, publication, live-home mutation, or broad config migration

### DIFF
--- a/changes/1812.fixed.md
+++ b/changes/1812.fixed.md
@@ -1,0 +1,6 @@
+`astrid run` now honors `timeouts.run_idle_secs` as its no-message idle
+timeout and accepts `--idle-timeout-secs` for one invocation. The default
+remains 120 seconds, valid active-run messages reset the timer, values above
+one day are rejected, and timeout cleanup/disconnect runs before the command
+returns the established exit code 53. Capsule HTTP timeout controls remain
+independent. Closes #1812
--- a/changes/1812.fixed.md
+++ b/changes/1812.fixed.md
@@ -1,6 +1,11 @@
-`astrid run` now honors `timeouts.run_idle_secs` as its no-message idle
-timeout and accepts `--idle-timeout-secs` for one invocation. The default
-remains 120 seconds, valid active-run messages reset the timer, values above
-one day are rejected, and timeout cleanup/disconnect runs before the command
-returns the established exit code 53. Capsule HTTP timeout controls remain
-independent. Closes #1812
+`astrid run` now loads its idle timeout from isolated pre-mount client
+configuration instead of the broad runtime config. `ASTRID_CLIENT_CONFIG_PATH`
+selects an explicit private file; the canonical AOS path is
+`~/.aos/etc/astrid/client.toml`. The CLI override remains highest precedence,
+the default remains 120 seconds, and values above one day are rejected.
+
+Only messages correlated to the active run reset the idle deadline. This
+includes session-bearing `agent.v1.stream.delta` payloads and excludes foreign
+same-principal approval/control frames. Timeout and disconnect sends are
+bounded, and the command still returns exit code 53 at the CLI boundary.
+Capsule HTTP timeout controls remain independent. Closes #1812 and #1816
--- a/crates/astrid-cli/src/commands/headless.rs
+++ b/crates/astrid-cli/src/commands/headless.rs
@@ -2,10 +2,77 @@
 
 use std::io::IsTerminal;
 
+use std::future::Future;
+use std::time::Duration;
+
 use anyhow::{Context, Result};
+use thiserror::Error;
 
 use super::daemon;
 use crate::{formatter, socket_client, tui};
+
+/// The established headless timeout exit code.
+pub(crate) const TIMEOUT_EXIT_CODE: u8 = 53;
+const SUCCESS_EXIT_CODE: u8 = 0;
+/// A hard ceiling that permits long cold starts but rejects unbounded waits.
+pub(crate) const MAX_RUN_IDLE_TIMEOUT_SECS: u64 = 86_400;
+
+/// Why response collection stopped without a terminal response.
+#[derive(Debug, Error)]
+pub(crate) enum HeadlessError {
+    /// No active-run message arrived within the configured idle budget.
+    #[error("timed out waiting for response after {timeout_secs}s idle")]
+    IdleTimeout {
+        /// The configured idle budget in whole seconds.
+        timeout_secs: u64,
+    },
+    /// The daemon connection failed while collecting the response.
+    #[error(transparent)]
+    Read(#[from] anyhow::Error),
+}
+
+/// Validate and convert a whole-second run idle timeout.
+///
+/// # Errors
+/// Returns an error for zero or values above the one-day operational ceiling.
+pub(crate) fn idle_timeout(timeout_secs: u64) -> Result<Duration> {
+    if timeout_secs == 0 {
+        anyhow::bail!("idle timeout must be greater than 0 seconds");
+    }
+    if timeout_secs > MAX_RUN_IDLE_TIMEOUT_SECS {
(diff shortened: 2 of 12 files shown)


joshuajbouw:
Not planned. HQ kept draft PR #1815 as the sole #1812/#1816 vehicle.

#1815 now points at exact successor `1e725a41e625e6bd7ed8b33c5d3eb2de1360ff9b` after a normal fast-forward of `fix/run-idle-timeout-1812`. This duplicate draft and branch `codex/1812-client-idle-timeout` are preserved as evidence. The branch is not deleted. No merge, tag, or publication.

